PR #11, opened 2026-05-16
Drop agent terminology from mail-memory internals

alexanderguy:
## Summary

- Collapse the parallel `#agentMailboxes` / `#cryptoProviders` maps in `InMemoryTransport` into a single `Map<string, MailboxEntry>` whose entry owns the crypto provider, making the registration invariant structural rather than convention-enforced.
- Drop the `eslint-disable @typescript-eslint/no-non-null-assertion` banner from `send.ts` and replace its `Map.get()!` patterns with explicit `undefined` checks.
- Rename the public API away from agent-shaped names: `registerAgent` → `register`, `unregisterAgent` → `unregister`, `getTransportForAgent` → `getTransportFor`, and the `deliver` parameter from `agentAddress` to `address`. Update every caller atomically.

## Changes

- `packages/mail-memory/src/mailbox.ts`: `AgentMailboxEntry` → `MailboxEntry` (now carries `crypto`); `createAgentEntry` → `createEntry(crypto)`.
- `packages/mail-memory/src/send.ts`: `executeSend` takes a single `entries` map; eslint-disable banner removed; descriptive `throw` paths replace `!` assertions.
- `packages/mail-memory/src/fetch.ts`: `fetchFull` and `verifyMessageSignature` take a `getCrypto: (address) => CryptoProvider | undefined` callback instead of the crypto map, keeping `fetch.ts` from learning about the broader entry shape.
- `packages/mail-memory/src/transport.ts`: `InMemoryTransport` holds one `#entries` map; `AgentMessageTransport` renamed to `ScopedMessageTransport`; public methods renamed and "agent" wording in doc comments and error strings scrubbed.
- `packages/mail-memory/src/index.ts`: docstring example updated to the new method names.
- `packages/mail-memory/src/index.test.ts`: three new regression tests in a `registration lifecycle` block that lock in the single-entry invariant (unregister-then-re-register, send-after-deregister, `fetchFull` returns `"unknown"` after sender removal); all existing call sites updated.
- `apps/sidecar/src/session-manager.ts`, `apps/sidecar/src/ws-client.test.ts`, `bin/posix-demo.ts`, `bin/ring-demo.ts`: consumer call sites updated to the renamed API.

## Out of scope

- The `interchange-agent-id` MIME header and `interchangeAgentId` fields owned by `@interchange/mime`.
- The `MessageTransport` interface in `@interchange/types/runtime` (still declares `agentAddress` on `subscribe`/`unsubscribe`). The local stubs were renamed to `_subscriberAddress`; the interface itself is a separate cleanup.
- The `session-manager.ts` local variable `agentAddress` — that is the harness's concept, not mail-memory's.

## Test plan

- [x] `bun run check`
- [x] `bun run lint`
- [x] `bun run test` (1119 tests pass, +3 new)
- [x] `grep -rni "agent" packages/mail-memory/src/` returns only the three out-of-scope `interchange-agent-id` references.

Closes INTR-48

### DIFF
--- a/apps/sidecar/src/session-manager.ts
+++ b/apps/sidecar/src/session-manager.ts
@@ -311,8 +311,8 @@ export function createSessionManager(
     try {
       const crypto = createNodeCrypto(keyPair);
 
-      transport.registerAgent(agentAddress, crypto);
-      const agentTransport = transport.getTransportForAgent(agentAddress);
+      transport.register(agentAddress, crypto);
+      const agentTransport = transport.getTransportFor(agentAddress);
 
       const sessionId = agentConfig.sessionId;
 
@@ -388,7 +388,7 @@ export function createSessionManager(
       mailStores.delete(agentAddress);
       mailCommitQueues.delete(agentAddress);
       try {
-        transport.unregisterAgent(agentAddress);
+        transport.unregister(agentAddress);
       } catch (cleanupErr) {
         // Best-effort cleanup; don't mask the original error.
         logger.error`Failed to unregister transport for ${agentAddress}: ${String(cleanupErr)}`;
@@ -413,7 +413,7 @@ export function createSessionManager(
     await drainMailQueue(agentAddress);
     sessions.delete(agentAddress);
     mailStores.delete(agentAddress);
-    transport.unregisterAgent(agentAddress);
+    transport.unregister(agentAddress);
     logger.info`Stopped session for ${agentAddress}`;
   }
 
@@ -435,7 +435,7 @@ export function createSessionManager(
     await drainMailQueue(agentAddress);
     sessions.delete(agentAddress);
     mailStores.delete(agentAddress);
-    transport.unregisterAgent(agentAddress);
+    transport.unregister(agentAddress);
     logger.info`Aborted agent ${agentAddress}: ${reason}`;
   }
 
--- a/apps/sidecar/src/ws-client.test.ts
+++ b/apps/sidecar/src/ws-client.test.ts
@@ -463,7 +463,7 @@ describe("sidecar↔hub integration", () => {
       "@interchange/crypto-node"
     );
     const kp = await generateKeyPair();
-    transport.registerAgent("agent-1@test.interchange", createNodeCrypto(kp));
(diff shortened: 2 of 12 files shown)


alexanderguy:
Code review notes.

Two commits, scoped to `packages/mail-memory` plus its consumers and the `MessageTransport` interface in `@interchange/types`:

- `666f9e8` collapses the parallel address-keyed maps inside `InMemoryTransport` into a single `Map<string, AddressEntry>` whose entry owns the `CryptoProvider`. Drops the file-level `eslint-disable @typescript-eslint/no-non-null-assertion` from `send.ts` in favor of explicit `undefined` checks that throw with descriptive errors. `fetch.ts` narrows to a `getCrypto: (addr) => CryptoProvider | undefined` callback so it only depends on what it actually needs for signature verification. Renames the `subscribe`/`unsubscribe` parameter on the `MessageTransport` interface (and the corresponding line in `docs/MESSAGE.md`) from `agentAddress` to `subscriberAddress`, aligning the interface with the local stub names. Adds a `registration lifecycle` test block exercising unregister-then-re-register, send-after-deregister, and the `fetchFull` returns-`"unknown"` path.
- `4bc72fe` renames the public API (`registerAgent` → `register`, `unregisterAgent` → `unregister`, `getTransportForAgent` → `getTransportFor`, the `deliver(agentAddress, ...)` parameter to `address`), scrubs the remaining `agent` wording from doc comments and error strings, and updates every caller atomically — `apps/sidecar/src/{session-manager,ws-client.test}.ts`, `bin/{posix-demo,ring-demo}.ts`, plus the package's own tests and docstring example.

Build pipeline (`bun run check && bun run lint && bun run test`) is clean: 1110 unit tests + 9 integration tests pass.

Scope discipline observed:

- The `interchange-agent-id` MIME header and `interchangeAgentId` field stay — wire format owned by `@interchange/mime`.
- `session-manager.ts`'s own local variable `agentAddress` is the harness's concept and is left alone.
- The `move()` `eslint-disable-next-line` on the index-bounded `!` assertion in `transport.ts` is narrowly scoped and out of scope here.

PR #12, opened 2026-05-16
Introduce a Makefile and expand bin/check-env

alexanderguy:
## Summary

- Add a top-level `Makefile` matching the verbs used in the sibling `faremeter/faremeter` tree (`all`, `build`, `lint`, `test`, `format`, `docs`, `clean`), adapted to this tree's bun + `tsc -b` tooling. Each target delegates to the corresponding `bun run` script so `package.json` remains the single source of truth.
- Expand `bin/check-env` to verify `bun` is on `PATH` at the version required by `DEV.md` and that `node_modules` exists, in addition to the existing git-hooks check.
- The `Makefile`'s `include .env-checked` pattern forces `bin/check-env` to run before any target, so prerequisites are validated up front.

Linear: [INTR-57](https://linear.app/abklabs/issue/INTR-57/introduce-makefile-and-expand-bincheck-env)

## Test plan

- [ ] `make all` succeeds on a healthy tree
- [ ] `make lint`, `make build`, `make test`, `make format`, `make docs` each run their respective `bun run` script
- [ ] `bin/check-env` fails with an actionable message when run outside a git repo, when git hooks aren't configured, when `bun` is missing or too old, and when `node_modules` is missing
- [ ] `make clean` removes `.env-checked`, `.eslintcache`, non-`node_modules` `tsconfig.tsbuildinfo` files, and non-`node_modules` `dist/` directories

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,32 @@
+export PATH := $(PWD)/bin:$(PATH)
+
+all: lint build test
+
+build: FORCE
+	bun run check
+
+lint: FORCE
+	bun run lint
+
+test: FORCE
+	bun run test
+
+format: FORCE
+	bun run format
+
+docs: FORCE
+	bun run docs
+
+clean:
+	rm -f .env-checked .eslintcache
+	find . -type f -name tsconfig.tsbuildinfo -a ! -path '*/node_modules/*' | xargs rm -f
+	find . -type d -name dist -a ! -path '*/node_modules/*' | xargs rm -rf
+
+.env-checked: bin/check-env
+	./bin/check-env
+	touch .env-checked
+
+include .env-checked
+
+.PHONY: all build lint test format docs clean
+FORCE:
--- a/bin/check-env
+++ b/bin/check-env
@@ -3,15 +3,35 @@
 opsh::version::require v0.7.0
 lib::import step-runner
 
+REQUIRED_BUN_VERSION=1.2.0
+
 explain-githooks() {
 	log::fatal "Please configure your git hooks: git config core.hooksPath .githooks"
 }
 
-step::10::check_git_hooks() {
+step::10::check_git_repo() {
 	git rev-parse --git-dir >/dev/null 2>&1 || log::fatal "Not inside a git repository"
+}
+
+step::20::check_git_hooks() {
 	hookspath=$(git config core.hooksPath) || explain-githooks
 	[[ $hookspath = .githooks ]] || explain-githooks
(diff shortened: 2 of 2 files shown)


alexanderguy:
## Code review

**Scope:** 2 commits, 2 files, +53 / -1.

```
7173adc Introduce a Makefile mirroring the faremeter tree's verbs
35d1887 Expand bin/check-env to verify bun and node_modules
```

Each commit touches exactly one file and its message accurately summarises the diff.

### Makefile

The structure is clean: every non-`clean` target is a thin wrapper around a `bun run` script, with `package.json` owning the actual command definitions. The `include .env-checked` pattern correctly forces `bin/check-env` to run before any target, and `clean` deletes the marker so the next invocation re-validates.

A couple of small observations, neither blocking:

- All non-pattern targets are declared both `.PHONY` and given a `FORCE:` dependency. The two mechanisms produce the same effect on these targets, so one is redundant. The reference faremeter Makefile uses `FORCE` because it has pattern targets like `packages/%` that `.PHONY` can't cover cleanly; this Makefile has none, so the `FORCE` dependencies on `build`, `lint`, etc. are vestigial. Worth keeping for symmetry with the sibling tree, but worth knowing they aren't doing work.
- `all: lint build test` orders lint before build. AGENTS.md describes the verification pipeline as `check && lint && test`. Mostly cosmetic — all three must pass regardless — but lint-first does surface formatting issues faster than tsc.

### bin/check-env

The four steps (`check_git_repo`, `check_git_hooks`, `check_bun`, `check_node_modules`) each do one thing and produce an actionable `log::fatal` message on failure. `REQUIRED_BUN_VERSION` lives at the top as a constant, which makes future bumps obvious. Using opsh's `semver::test` is the right call — bun stable releases use plain `MAJOR.MINOR.PATCH` and parse correctly.

One pre-existing limitation worth noting (not introduced by this PR): the script assumes CWD is the repo root, since `[[ -d node_modules ]]` and the `hookspath` check are both CWD-relative. Running `bin/check-env` from a subdirectory would falsely fail. The Makefile invokes it from the repo root so this is fine in practice; just something to be aware of if the script gets called from anywhere else.

### Things considered and dismissed

- **`xargs rm` on empty input in `clean`** — verified locally that macOS `xargs` skips invocation when input is empty, matching GNU's `-r` default. No spurious failure on a fresh tree.
- **`bun --version` prerelease handling** — bun stable releases are plain semver; `semver::test` handles them correctly. Not a concern in practice.

### Overall

Small, focused, verified end-to-end (pre-commit lint passes on both commits, `bin/check-env` failure paths produce useful messages). Ready to merge.

PR #13, opened 2026-05-16
Route runInference through an injected Dependencies parameter

alexanderguy:
## Summary

- Adds a `Dependencies` parameter (`fetch` + optional `[HarnessId]` symbol tag) threaded through `runInference` and `createReactor`, so the streaming harness reads `deps.fetch` instead of `globalThis.fetch`.
- Production callers wire `createDefaultDependencies()` at the reactor surface. Tests can substitute a stub `fetch` without monkey-patching globals.
- A guard at the top of `runInference` throws a plain `Error` if `deps.fetch` isn't callable, so a programmer bug never gets misclassified as a retryable transport error.
- `ProviderConfig` (the persisted arktype schema) stays unchanged.

## Test plan

- [x] `bun run check` (full project TypeScript graph)
- [x] `bun run lint` (Prettier, ESLint, API docs freshness)
- [x] `bun run test` — 1124 unit + 9 integration, all pass
- [x] New `packages/inference/src/harness.test.ts` covers: happy path (`deps.fetch` invoked, `globalThis.fetch` never reached), transport-error classification (`retryable` + preserved cause message), three crash-loudly cases (undefined `deps.fetch`, non-function `deps.fetch`, omitted `deps`), `createDefaultDependencies` binding contract, and the reflective-exposure claims the JSDoc makes about `[HarnessId]`.

Closes INTR-58

### DIFF
--- a/packages/harness/src/harness.ts
+++ b/packages/harness/src/harness.ts
@@ -22,6 +22,7 @@ import {
   createReactor,
   createAuditCollector,
   createAuthzExtension,
+  createDefaultDependencies,
   createSizeCapTransform,
 } from "@interchange/inference";
 import type {
@@ -410,6 +411,7 @@ export function createHarness(config: HarnessConfig): Harness {
     toolRunner: combinedRunner,
     contextStore,
     onEvent: handleEvent,
+    deps: createDefaultDependencies(),
     beforeToolExtensions,
     toolResultTransforms: [sizeCapTransform],
   };
--- a/packages/inference/src/harness.test.ts
+++ b/packages/inference/src/harness.test.ts
@@ -1,0 +1,302 @@
+import { describe, test, expect } from "bun:test";
+
+import {
+  createDefaultDependencies,
+  HarnessId,
+  runInference,
+  type Dependencies,
+  type InferenceHarnessOptions,
+} from "./harness";
+import type {
+  ConversationTurn,
+  InferenceEvent,
+} from "@interchange/types/runtime";
+
+const PROVIDER_CONFIG = {
+  provider: "anthropic" as const,
+  baseURL: "https://api.anthropic.com",
+  apiKey: "test",
+};
+
+function userTurn(text: string): ConversationTurn {
+  return {
+    role: "user",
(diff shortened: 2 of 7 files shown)


alexanderguy:
## Code review

**Base:** `main`
**Scope:** 1 commit, 7 files, +363/-5

Single-purpose change. The implementation matches the spec, tests pin the contract, conventions followed, scope is tight. No blocking, should-fix, or nit findings.

### Walked through and clean

- **Spec contract** — Every clause in the Inference Harness Plan's "HTTP test seam" section is enforced by code and pinned by a test: narrow `Dependencies.fetch` shape (not `typeof globalThis.fetch`), required `deps` parameter, throw-before-yield contract, JSDoc-documented symbol-leak vectors with `JSON.stringify` correctly noted as safe.
- **Issue outcomes** — All eight INTR-58 outcome bullets satisfied; `ProviderConfig` literally untouched.
- **Convention compliance** — Imports ordered external → internal → relative; named exports only; no `as Type` (the three test-side `as unknown as` casts each have eslint-disable + justification per `CONVENTIONS.md`); no `!` non-null; defensive coding (no fallbacks, errors surface); `HarnessId` follows the project's "ID is camelCase" rule.
- **Test coverage** — Pins the contract, not the implementation: plain `Error` (not subclass), no events yielded on throw, error message contains the field name and observed `typeof`, transport-error path preserves `category: "retryable"` and cause message, three symbol-key reflective probes pin the JSDoc claims directly with a proper string-keyed control to prove the assertions aren't tautological.
- **Cross-file consistency** — Every `createReactor` and `runInference` caller in the tree passes `deps`. The only direct `globalThis.fetch` reference left in production code is inside `createDefaultDependencies` itself (the intended single binding site).
- **Scope discipline** — No drive-by changes, no unrelated reformatting. The `packages/harness/src/harness.ts` and `test/integration/transform-cutover.test.ts` edits are forced consequences of the type change.
- **Commit-message coherence** — The diff matches the stated purpose. Both behaviors the commit ships (routing through `deps.fetch` and the up-front guard) are called out in the body.

Verdict: ship.

PR #14, opened 2026-05-17
Add a virtual clock to @interchange/inference-testing

alexanderguy:
## Summary

- Introduces the `@interchange/inference-testing` package skeleton, wired into the root `tsconfig` project graph.
- Adds a deterministic virtual clock implementing the locked virtual-clock contract: absolute-time `schedule`, monotonic-seq sub-tick ordering, bounded microtask quiescence (default budget 256), and a wall-clock watchdog (default 250 ms, `Infinity` opts out).
- Sync callback throws propagate via an optional `onSyncCallbackError` hook the harness will wire in later; defaults to a no-op.

## Changes

- `packages/inference-testing/` — new package, no runtime or peer dependencies.
- `packages/inference-testing/src/clock.ts` — `Clock` interface, `createClock` factory, `ClockOverrunError`, `ClockWallClockOverrunError`, and a bundled binary min-heap.
- `packages/inference-testing/src/clock-internal.ts` — `@internal` test-only `createClockWithSeq` seam for the monotonic-seq overflow assertion; not re-exported from the package barrel.
- `packages/inference-testing/src/clock.test.ts`, `clock-boundary.test.ts` — unit tests covering ordering, quiescence, overrun, defensive validation, hook propagation, seq overflow, and boundary semantics.
- `tsconfig.json` — adds `packages/inference-testing` to the project references array.

## Testing

- [x] `make all` (lint + `tsc -b` + tests) passes against the full repo.
- [x] New tests: 30 in this package; the broader suite remains green at 1150 + 9 integration.

Closes INTR-59


alexanderguy:
## Code review

### Commit-message coherence

Both commits' diffs match their messages. No undescribed drive-by changes, no out-of-scope files.

### Findings

**Blocking:** none.
**Warnings:** none.

**Nits (take or leave):**

- None outstanding; the prior review nits have been folded in:
  - `createClockInternal` carries an `@internal TEST-ONLY` JSDoc explaining the production vs. test seam.
  - `fireOne`'s `state.activity += 1` carries a comment noting it is defensive against re-entrant drains.
  - `onSyncCallbackError` has TSDoc describing the future-harness stream-cleanup use case and the throwing-hook wrapping behavior.
  - Heap code factored to `heapAt`/`heapSwap` helpers, replacing the inline invariant-guard noise.

### Convention compliance

- TypeScript strict, no `any`, no `as Type`, `import type` used where applicable; `verbatimModuleSyntax` honored.
- Naming follows project conventions: `createClock` factory, `Clock` interface, `ClockOverrunError`/`ClockWallClockOverrunError` Error subclasses, camelCase methods, acronyms preserved.
- No `console.*`. No real timers in the implementation; only `queueMicrotask` and `performance.now()`.
- Zero new dependencies; the package stays zero-dep.
- Defensive coding throughout: past-time `schedule`/`advanceTo` throw, non-finite `virtualMs` throws, invalid `microtaskBudget`/`wallClockBudgetMs` throws, non-function `fn`/`hook` throws, `monotonicSeq` overflow throws.
- Tests via `bun:test`, co-located as `*.test.ts`.
- Diff scope is clean: only `packages/inference-testing/**` plus a single line added to the root `tsconfig.json` references array.

### Verdict

Ship it. The branch implements the locked spec end-to-end; the API matches exactly; the semantics are correct on every checked invariant; the test suite would catch each spec requirement if regressed.

PR #16, opened 2026-05-17
Extract reactor-assembly helper and migrate harness onto it

alexanderguy:
## Summary

- Add `createReactorAssembly` in `@interchange/inference`, the canonical composition of the standard reactor wiring (audit collector, authz extension, before-tool extensions, size-cap transform default, blob reader, audit-flush lifecycle hooks).
- Migrate `createHarness` onto the helper. Reactor wiring collapses from ~90 inline lines into one call. Behavior, the public `Harness` type, and the existing harness tests are unchanged.
- Unblocks INTR-51 (in-process agent package), which will reuse the helper without depending on mail transport.

## Changes

- `packages/inference/src/assembly.ts` (new): `createReactorAssembly`, `ReactorAssemblyConfig`, `ReactorAssembly`. Composes a size-cap tool-result transform (prepended, default 10_000 chars), an authz before-tool extension when `authorize` is supplied, an audit collector when `auditStore` is supplied, chained `afterCheckpoint`/`onShutdown` hooks that flush audit before the caller's hook runs, and a `BlobReader` over the supplied context store. Does NOT wrap the context store — that concern belongs to the caller.
- `packages/inference/src/assembly.test.ts` (new): 12 unit tests covering composition order, audit-flush ordering, blob reader wiring, the absent-audit and absent-authz paths, and the `composedOnEvent` interception that feeds the audit collector.
- `packages/inference/src/index.ts`: re-export the helper and its types.
- `packages/harness/src/harness.ts`: replace the inline reactor-assembly wiring with one `createReactorAssembly` call. Rename the local connector-state-wrapped store from `contextStore` to `wrappedStore` to make the layering explicit. Remove the now-redundant `auditCollector.onEvent(event)` feed in `handleEvent` (the helper composes that path). Rework `setProviderConfig` to mutate a locally-cloned `providerConfig` object in place since the helper now owns the `reactorConfig` object reference; the reactor's lazy read of `config.providerConfig` per inference call observes the mutations.

## Why context-store wrapping stays in harness

The wrapper at `packages/harness/src/harness.ts:163-217` is glue between harness-owned mutable mail state (`connectorThreadRoot`, `connectorLastMessageId`, `connectorReplyTo`, `connectorSubject`) and the store's per-cycle `writeMetadata` rhythm. Moving it into the helper means the helper owns the mail state, at which point it stops being reactor-assembly. INTR-51 will produce a real second consumer; that's the right moment to factor a shared decorator against two real shapes, not one and a guess.

## Test plan

- [x] `make all` green on the branch (lint, `tsc -b`, 1302 main + 9 sidecar/inference integration tests).
- [x] `packages/harness/src/harness.test.ts` and `packages/harness/src/deploy-tree.test.ts` byte-identical to `origin/main` (hard gate).
- [x] 12 new tests in `packages/inference/src/assembly.test.ts` cover the composition contract end-to-end.
- [x] No drive-by changes outside the four expected files.

Closes INTR-49

### DIFF
--- a/packages/harness/src/harness.ts
+++ b/packages/harness/src/harness.ts
@@ -18,21 +18,10 @@
 // (ARCHITECTURE.md § Agent Harness, INFERENCE.md § Relationship to Harness)
 
 import { getLogger } from "@interchange/log";
-import {
-  createReactor,
-  createAuditCollector,
-  createAuthzExtension,
-  createDefaultDependencies,
-  createSizeCapTransform,
-} from "@interchange/inference";
-import type {
-  Reactor,
-  AuditCollector,
-  ReactorEmittedEvent,
-} from "@interchange/inference";
+import { createReactorAssembly } from "@interchange/inference";
+import type { ReactorEmittedEvent } from "@interchange/inference";
 import {
   ProviderConfig,
-  createBlobReader,
   type BlobReader,
   type ContextStore,
   type ConnectorThreadState,
@@ -126,29 +115,10 @@ export function createHarness(config: HarnessConfig): Harness {
 
   const sessionId = crypto.randomUUID();
 
-  // -------------------------------------------------------------------------
-  // Audit collector: correlates tool events with authz decisions.
-  // -------------------------------------------------------------------------
-
   const auditStore = config.auditStore;
-  const auditCollector: AuditCollector | undefined =
-    auditStore !== undefined ? createAuditCollector(sessionId) : undefined;
 
   const accumulatedErrors: ErrorRecord[] = [];
   let errorSeq = 0;
-
-  const authzExtension =
-    config.authorize !== undefined
-      ? createAuthzExtension({
-          authorize: config.authorize,
-          onDecision: (d) => auditCollector?.onDecision(d),
-        })
-      : undefined;
-
-  const beforeToolExtensions = [
-    ...(authzExtension !== undefined ? [authzExtension] : []),
-    ...(config.beforeToolExtensions ?? []),
-  ];
 
   // -------------------------------------------------------------------------
   // Connector state: track which thread(s) this reactor owns.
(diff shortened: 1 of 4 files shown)


alexanderguy:

Reviewed `origin/main..HEAD` (two commits, four files, +1020/-86).

### Verified

- `git diff origin/main -- packages/harness/src/{harness.test.ts,deploy-tree.test.ts}` empty — the existing harness tests are byte-identical to main.
- Only the four expected files modified; no drive-by changes elsewhere.
- `make all` green (lint + `tsc -b` + 1302 unit + 9 integration tests).
- Helper composition matches `createHarness` today: `composedOnEvent` predicate (`event.type !== "message.received"`), authz prepended to `beforeToolExtensions`, size-cap prepended to `toolResultTransforms` (default 10_000), audit flush ordering (helper flush → caller's hook), conditional `afterCheckpoint`/`onShutdown` only when audit is wired, `deps` always defined (default `createDefaultDependencies()`).
- `setProviderConfig` clone-and-mutate (`packages/harness/src/harness.ts:352`, `:449-465`) is safe: `ProviderConfig` (`packages/types/src/runtime.ts:1443`) is flat (four string fields), so a shallow clone is sufficient; `delete providerConfig.model` is the correct operator under `exactOptionalPropertyTypes`; the reactor's lazy read of `config.providerConfig` per inference call picks up the mutated fields.
- All 12 plan-required tests present in `packages/inference/src/assembly.test.ts`, each asserting on observable behavior (blob spills, audit records, trace ordering) rather than internal state.
- TSDoc on `createReactorAssembly`, `ReactorAssemblyConfig`, `ReactorAssembly` designates the helper as the canonical reactor-assembly path for future consumers.
- Conventions: named exports only, `import type` for type-only imports, no `any`, no unsafe `as` casts, no `console`, hierarchical logger.

### Findings

None blocking.

### Notes

- `packages/inference/src/assembly.ts:183-189` — `flushAudit` guards on `auditCollector === undefined || auditStore === undefined`, but the only call sites are themselves gated on `auditCollector !== undefined`. Dead defense in depth. Left as-is; removing it would couple the helper's internal structural assumptions across function bodies.
- `packages/inference/src/assembly.ts:32` — imports `createDefaultDependencies` and `Dependencies` from `./harness` (a pre-existing file inside the inference package, unrelated to `@interchange/harness`). The shared filename is confusing but pre-dates this branch.

PR #18, opened 2026-05-18
Add @interchange/agent and coding-agent example

alexanderguy:
## Summary

- New `@interchange/agent` package: in-process agent runtime sitting on `createReactorAssembly` from `@interchange/inference`. Peer to `@interchange/harness` (the mail-transport-bound consumer); the agent is the code-driven `send`/`stream`/`history` consumer.
- New `examples/coding-agent/`: reference consumer wiring `createAgent` with `@interchange/tools-posix` + `@interchange/tools-lsp`, persisting to `tmp/coding-agent/context/` so re-running resumes prior conversations. CLI is verified end-to-end through `@interchange/inference-testing` (parses argv, drives the reactor, prints reply, exit code; resume-from-crash by reopening the same `contextDir`).
- `examples/` established as a first-class workspace alongside `apps/` and `packages/` (gated by `make all`, listed in root `tsconfig.json` references, documented in `CONVENTIONS.md`).
- Two small ergonomic additions to `@interchange/inference-testing` driven by writing the integration tests: `scenario.replyOnce(provider, opts)` collapses the create-stream + build-chunks + enqueue + register-matcher ritual into one call, and `scenario.lastRequest()` returns a clone of the most-recently matched `Request` so tests can assert on the body shape without instrumenting a capture closure inside the (sync-only) matcher predicate.

Public agent surface: `createAgent`, `send`, `stream`, `deliver`, `close`, `setProvider`, `history`, `checkpoints`, `readAt`, `blobReader`, plus tool sugar (`tool`, `stringTool`, `fromToolRunner`).

Notable design points:
- **Singleton-per-contextDir**: runtime check via `AgentInUseError` (path-resolved lock; same `contextStore` instance passed by the caller bypasses the lock).
- **FIFO send queue** with configurable `sendQueueMax` (default 16); per-send `AbortSignal` removes queued items or rejects in-flight callers; overflow throws `SendQueueFullError` synchronously.
- **Bounded `stream()` fan-out** per consumer with `streamBufferMax` (default 1024); overflow throws `StreamBackpressureError` into the affected iterator only.
- **`setProvider` rotates everything**, including `model`. The agent wraps the director's `capabilities.infer(model, ...)` so the live model is substituted on every call — works with the default director and any caller-supplied director that simply relays its captured model.
- **`close()` waits for shutdown** (audit flush + in-flight commits) before releasing the contextDir lock, bounded by `closeTimeoutMs` (default 5s) as a backstop. Fallback resolution on `reactor.done` for paths where `onShutdown` never fires.
- Non-fatal `reactor.error` events do not terminate the active send — the cycle may still produce a `connector.reply`.

## Test plan

- [x] `make all` clean: **1429 tests** passing across 77 files, no regressions in existing 1300+ tests.
- [x] Per-commit `tsc -b` clean for all 17 commits.
- [x] Per-commit `bun test packages/agent tests/agent` passes for every commit; test count grows monotonically.
- [x] Lifecycle tests against real isogit store: singleton lock, close+reopen on same dir, idempotent close, post-close error contract.
- [x] Send-flow integration via `@interchange/inference-testing`: round-trip send/reply, history persistence across reopen, `readAt` against recorded checkpoint, close-while-pending drainage with `AgentClosedError`, `setProvider` credential rotation, `setProvider` model rotation (asserts on-wire model name via `lastRequest()`), abort-mid-flight, `SendQueueFullError` past cap.
- [x] **CLI end-to-end** via the harness: argv parsing, agent construction, send, stream pump, close, exit codes; missing-API-key and empty-prompt failure paths; resume-from-crash (two invocations on the same `contextDir` produce ≥4 history turns the second run can see).

Closes INTR-51

### DIFF
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -29,8 +29,11 @@ This is a bun monorepo. When developing new TypeScript code:
 
 - **Applications** go in `apps/`
 - **Shared libraries** go in `packages/`
+- **Reference consumers** go in `examples/`
 
 Do not create standalone TypeScript files in the repository root.
+
+`examples/` is for runnable reference consumers that demonstrate how to use a package's public API. Examples are first-class workspace members: they have their own `package.json`, are listed in the root `tsconfig.json` `references` array, and are gated by `make all` (lint, type-check, test) just like packages and apps. They are not throwaway scratch code.
 
 ---
 
--- a/bun.lock
+++ b/bun.lock
@@ -73,6 +73,17 @@
         "vite": "^6.3.5",
       },
     },
+    "packages/agent": {
+      "name": "@interchange/agent",
+      "version": "0.0.0",
+      "dependencies": {
+        "@interchange/inference": "workspace:*",
+        "@interchange/mime": "workspace:*",
+        "@interchange/storage-isogit": "workspace:*",
+        "@interchange/types": "workspace:*",
+        "arktype": "catalog:",
+      },
+    },
     "packages/authz": {
       "name": "@interchange/authz",
       "version": "0.0.0",
@@ -424,6 +435,8 @@
     "@inquirer/figures": ["@inquirer/figures@1.0.15", "", {}, "sha512-t2IEY+unGHOzAaVM5Xx6DEWKeXlDDcNPeDyUpsRc6CUhBfU3VQOEl+Vssh7VNp1dR8MdUJBWhuObjXCsVpjN5g=="],
 
     "@inquirer/type": ["@inquirer/type@3.0.10", "", { "peerDependencies": { "@types/node": ">=18" }, "optionalPeers": ["@types/node"] }, "sha512-BvziSRxfz5Ov8ch0z/n3oijRSEcEsHnhggm4xFZe93DHcUCTlutlq9Ox4SVENAfcRD22UQq7T/atg9Wr3k09eA=="],
+
+    "@interchange/agent": ["@interchange/agent@workspace:packages/agent"],
 
     "@interchange/authz": ["@interchange/authz@workspace:packages/authz"],
 
--- a/bun.lock
+++ b/bun.lock
@@ -73,6 +73,16 @@
         "vite": "^6.3.5",
       },
     },
+    "examples/coding-agent": {
+      "name": "@interchange/example-coding-agent",
+      "version": "0.0.0",
+      "dependencies": {
+        "@interchange/agent": "workspace:*",
+        "@interchange/tools-lsp": "workspace:*",
+        "@interchange/tools-posix": "workspace:*",
+        "@interchange/types": "workspace:*",
+      },
+    },
     "packages/agent": {
       "name": "@interchange/agent",
       "version": "0.0.0",
@@ -444,6 +454,8 @@
     "@interchange/crypto-node": ["@interchange/crypto-node@workspace:packages/crypto-node"],
 
     "@interchange/db": ["@interchange/db@workspace:packages/db"],
+
+    "@interchange/example-coding-agent": ["@interchange/example-coding-agent@workspace:examples/coding-agent"],
 
     "@interchange/harness": ["@interchange/harness@workspace:packages/harness"],
 
--- a/bun.lock
+++ b/bun.lock
@@ -77,6 +77,7 @@
       "name": "@interchange/agent",
       "version": "0.0.0",
       "dependencies": {
+        "@interchange/harness": "workspace:*",
         "@interchange/inference": "workspace:*",
         "@interchange/mime": "workspace:*",
         "@interchange/storage-isogit": "workspace:*",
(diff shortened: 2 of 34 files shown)


alexanderguy:

The branch was reviewed against `origin/main` using the `code-review` skill, with `critique` and `greybeard` subagents as additional reviewers. **No outstanding major findings.** The notes below describe the state of the code as it stands, organized by where a reviewer is most likely to scrutinize.

### Correctness

- `handleEvent` only terminates the active send on a *fatal* `reactor.error`. Non-fatal errors are surfaced via `stream()` but the cycle may still produce a `connector.reply`.
- `close()` waits for the reactor's shutdown sequence (audit flush + in-flight commits) before releasing the singleton `contextDir` lock, so a subsequent `createAgent` on the same directory does not race background writers against the same `.git`. Bounded by `closeTimeoutMs` (default 5s) so a stuck reactor cannot deadlock the caller. `shutdownComplete` resolves from either the assembly's `onShutdown` hook or the reactor's terminal `reactor.done` event, so the wait does not burn the full timeout on paths where `onShutdown` never fires.
- `setProvider` rotates all of `provider`/`baseURL`/`apiKey`/`model`. The agent wraps the director the reactor sees so every `capabilities.infer(model, ...)` call substitutes the live model from `providerRegistry.active`. The default director and any caller-supplied director that simply relays its captured model both pick up the rotation on the next inference.
- `send()` is FIFO-queued. Concurrent calls are serialized; queue depth (active + pending) is bounded by `sendQueueMax` (default 16), overflow throws `SendQueueFullError` synchronously. Per-send `AbortSignal` removes queued items or rejects in-flight callers — when an in-flight send aborts, the underlying reactor cycle continues in the background so queue ordering against actual reactor cycles is preserved.
- `stream()` returns a bounded per-consumer iterator. Overflow throws `StreamBackpressureError` into the affected iterator only; other consumers are unaffected.
- `AgentInUseError` enforces the singleton-per-`contextDir` invariant via a process-wide path-resolved lock. The `contextStore` form of `createAgent` bypasses the lock (caller owns lifetime).

### Safety

- Errors surface rather than being swallowed: queue overflow is synchronous, abort signals reject the caller's promise, reactor-fatal errors propagate, stream backpressure poisons the affected iterator instead of dropping events.
- `close()` is idempotent and ordered: `closed = true` is set first, then the reactor is aborted, then the queue is drained with `AgentClosedError`, then stream consumers are closed, then shutdown is awaited, then the lock releases.
- The model-swap wrapper falls back to the initial validated model if `setProvider` is ever passed a config without `model`, rather than producing an inference call with undefined model.

### Scope

- Each commit's diff matches its message and is scoped to one logical change. The `buildInboundMessage` payload→content bug fix lives in the commit that introduced `buildInboundMessage` (not in the test commit that surfaced it). The `tool.test.ts` stub cleanup lives in the commit that introduced the stub (not in the example commit that happened to be where I noticed it).
- No drive-by lint fixes, formatting changes, or unrelated refactors in feature commits.
- The two `@interchange/inference-testing` additions (`scenario.replyOnce`, `scenario.lastRequest`) are in their own commit, separate from the test refactor that uses them.

### Build / test

- `make all`: **1429 tests passing across 77 files**. No regressions in the existing 1300+ tests.
- Per-commit `tsc -b` clean for all 17 commits.
- Per-commit tests pass for every commit; test count grows monotonically.
- Lifecycle tests against the real isogit store cover the singleton lock, close+reopen on same dir, idempotent close, and the post-close error contract for `deliver`/`setProvider`/`send`.
- Send-flow integration tests via `@interchange/inference-testing` cover the round-trip send/reply, history persistence across reopen, `readAt` against a recorded checkpoint, close-while-pending drainage with `AgentClosedError`, `setProvider` credential rotation, `setProvider` model rotation (asserts the on-wire `model` field via `scenario.lastRequest()`), abort-mid-flight, and `SendQueueFullError` past the configured cap.
- CLI end-to-end test drives the example's `main()` through the harness: argv parsing, agent construction, send, stream pump, close, exit codes; missing-API-key and empty-prompt failure paths; resume-from-crash (two invocations on the same `contextDir` produce ≥4 history turns the second run can see).

### Convention

- TypeScript strict mode, `verbatimModuleSyntax` `import type` discipline, no `as Type` assertions except where the existing `eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion` carve-outs apply (each annotated with a justification).
- arktype runtime validation at all boundaries that accept external input (provider configs, inbound messages).
- No `console.*`; the agent emits nothing on its own. The example CLI writes to `process.stdout` / `process.stderr` directly because that is the CLI's contract.
- Commit messages are plain English sentences, no prefix, no terminal punctuation. Subject lines under 72 characters; body lines wrapped.

### Known design trade-offs (documented in the code, not bugs)

- A second concurrent `close()` returns immediately while the first is still awaiting shutdown — caller cannot know from the second call alone whether the lock is yet released. Documented as best-effort.
- The `let sendQueue` forward declaration has a `prefer-const` eslint-disable: the closure ordering between `handleEvent`, `createReactorAssembly`, and `sendQueue` genuinely requires forward reference.
- `lifecycle.test.ts:close releases the lock` asserts post-close `history()` returns empty but does not prove every background writer has quiesced; the deeper guarantee is provided by `close()`'s shutdown await, not by that test.

PR #22, opened 2026-05-19
Make the session plane composable via an event-emitter API

alexanderguy:
## Summary

- The sidecar router's receiver-dispatch surface is exposed as a typed event emitter; alternate hosts can subscribe to the events they care about without re-supplying every callback.
- The hub's inline session orchestration becomes `createHubSessionOrchestrator(deps)` in the hub package, driven by a narrow `HubSessionRouterFacade` so it can be unit-tested without spinning up a router.
- Wire-layer queries (public key, deploy ref, mail persist, state pack ingest) live as a single `lookups` struct on the router config.

## Changes

- `packages/hub/src/ws/sidecar-events.ts`: typed `SidecarEventEmitter` with `emit` (notification, per-listener try/catch) and `emitAndAwait` (sequential, rethrow-first) semantics; `SidecarLookups` struct for queries.
- `packages/hub/src/ws/sidecar-handler.ts`: wire layer emits events and reads lookups; legacy callback fields gone. `SidecarRouter` now exposes `events`.
- `packages/hub/src/hub-session-lookups.ts`: `createHubSessionLookups({ db, agentRepoStore })` returning `Required<SidecarLookups>`.
- `packages/hub/src/hub-session-orchestrator.ts`: `createHubSessionOrchestrator(deps)` subscribing to events through a narrow router facade; returns `{ dispose }`.
- `packages/hub/src/hub-session-orchestrator.test.ts`: 11 unit tests covering each subscriber, the reconnect failure paths, and disposal.
- `apps/hub/src/index.ts`: entry point rewritten — constructs lookups, router, event collectors, and orchestrator in order.
- `packages/types/src/hex.ts`: shared `hexEncode` / `hexDecode` consolidating nine local copies across hub, sidecar, and tests.

Migrated tests in `packages/hub/src/ws/sidecar-handler.test.ts`, `apps/sidecar/src/ws-client.test.ts`, and `tests/sidecar/deploy-flow.test.ts` to the new emitter + lookups surface.

## Testing

- [ ] \`make all\` (lint, type check, 1485 unit tests, 9 sidecar integration tests)
- [ ] Hub starts cleanly and registers a sidecar
- [ ] Reconnect with valid challenge → \`agent.reconnected\` listener runs: grants refreshed, providers pushed, status flipped to running, event collector restored
- [ ] Reconnect with stale deployRef → \`deploy.ref.stale\` listener re-packs and pushes
- [ ] Agent A → Agent B mail produces \`mail.delivered\` SSE events for sender (direction=outbound) and recipient (direction=inbound)

Closes INTR-65

### DIFF
--- a/apps/hub/src/index.ts
+++ b/apps/hub/src/index.ts
@@ -10,6 +10,7 @@ import {
   createSidecarRouter,
 } from "@interchange/hub";
 import { generateKeyPair } from "@interchange/crypto-node";
+import { hexEncode } from "@interchange/types";
 import { upgradeWebSocket, websocket } from "hono/bun";
 import { setup, getLogger } from "@interchange/log";
 
@@ -35,12 +36,6 @@ if (!hubDataDir) {
 
 const hubSigningKey = await generateKeyPair();
 log.info("Generated hub deploy signing key");
-
-function hexEncode(bytes: Uint8Array): string {
-  return Array.from(bytes)
-    .map((b) => b.toString(16).padStart(2, "0"))
-    .join("");
-}
 
 const agentRepoStore = createAgentRepoStore({
   dataDir: hubDataDir,
--- a/apps/hub/src/index.ts
+++ b/apps/hub/src/index.ts
@@ -1,25 +1,17 @@
-import { eq, and, isNull } from "drizzle-orm";
-import {
-  createDB,
-  createGrantStore,
-  resolveInstanceProviders,
-} from "@interchange/db";
-import { agentInstance, sessionMail } from "@interchange/db/schema";
+import { createDB, createGrantStore } from "@interchange/db";
 import {
   createAgentRepoStore,
   createApp,
   createAuth,
   createEventCollectorRegistry,
+  createHubSessionLookups,
+  createHubSessionOrchestrator,
   createSessionService,
   createSidecarRouter,
-  generateId,
 } from "@interchange/hub";
 import { generateKeyPair } from "@interchange/crypto-node";
-import { parseMailToEmail } from "@interchange/mime";
-import { parseInferenceEvent } from "@interchange/types/runtime";
 import { upgradeWebSocket, websocket } from "hono/bun";
 import { setup, getLogger } from "@interchange/log";
-import { type } from "arktype";
 
 await setup();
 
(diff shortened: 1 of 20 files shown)


alexanderguy:
## Self-Review (\`code-review\` skill)

### What I looked at

- **Async emit semantics.** Two-shape emitter: \`emit\` (notification, per-listener try/catch + log) for events where the wire layer doesn't depend on listener outcome; \`emitAndAwait\` (sequential, rethrow-first) for \`agent.reconnected\` and \`agent.deploy.ack\` where a listener throw must roll back wire state. \`listenerCount === 0\` short-circuits the awaited path so the unconfigured-handler case stays synchronous.
- **Reconnect rollback.** \`agent.reconnected\` is emitted per-address inside the verification loop; a throw rolls back that address only. The supersede check (\`connections.get(ws) !== conn\`) is preserved. Confirmed by the existing test where a listener for one address throws and other addresses still flush their queues.
- **Lookups vs events split.** Multi-subscriber notifications go through the emitter; one-handler-returns-a-value queries go through \`lookups\`. \`SidecarLookups\` is documented to call out that \`lookupDeployRef\` and the \`deploy.ref.stale\` event are paired by host convention, not enforced by the router.
- **Facade pattern.** \`HubSessionRouterFacade\` (four methods: \`sendGrantsUpdate\`, \`sendProvidersUpdate\`, \`sendPack\`, \`dispatchAgentEvent\`) is what the orchestrator imports — the full \`SidecarRouter\` is never referenced in \`hub-session-orchestrator.ts\`, so the unit tests need only a stub.
- **DB mock fidelity.** Orchestrator tests mock drizzle with \`as unknown as DB[\"db\"]\` and per-test canned data. The mock covers the methods exercised; the credential-resolution branch is left as a negative assertion (empty result → no \`sendProvidersUpdate\` call) rather than reproducing the full provider chain.
- **mail.delivered self-dispatch.** The \`persistMail\` lookup returns both the sender's outbound row and recipients' inbound rows; the orchestrator emits \`mail.delivered\` for each. The SSE consumer at \`packages/hub-client/src/transforms.ts:107\` keys off \`direction\` to render the outbound row as \`role: \"assistant\"\` — intentional, not a bug.
- **Hex helpers.** Five \`hexEncode\` and four \`hexDecode\` copies collapsed into \`@interchange/types\`. Validated decoder uses \"odd-length\"/\"non-hex characters\" wording to match the existing assertion in \`ws-client.test.ts\`.
- **Test migration completeness.** \`grep -rn 'function hexEncode|function hexDecode'\` returns only the canonical pair. Legacy \`on*\` callback names appear nowhere outside the typed event map's prose docs.

### Findings

Clean. No correctness, safety, scope, build/test, or convention issues outstanding.

PR #23, opened 2026-05-19
Make hub routes composable as DI factories, split createApp

alexanderguy:
## Summary

- Every hub route group is now exported as a dependency-injected factory (`createXRoutes(deps)`). Per-request middleware that consumed app-lifetime services (`resolveTenant`, `requireGrant`) becomes a factory of the same shape.
- The session-layer variables (db, grantStore, conditionRegistry, sessionService, sidecarRouter, eventCollectors) are gone from `AppEnv.Variables`; the setter middleware that populated them is deleted. The request context now carries only per-request facts.
- `createHubContextMiddleware` and `mountHubRoutes` are exported from the hub package. Third parties can stand up their own Hono application, mount the request logger and context middleware on their own terms, and bring in hub routes via `mountHubRoutes`. `createApp` is rebuilt as a thin convenience over the two helpers with an unchanged first-party signature.

## Test plan

- [x] \`make all\` (lint + \`tsc -b\` + tests) passes locally on a clean rebase from \`origin/main\`.
- [x] \`packages/hub/src/app.test.ts\` and \`packages/hub/src/routes/instances.test.ts\` continue to pass through the public \`createApp\` path.
- [x] A grep of \`c.get(\` across \`packages/hub/src\` returns only \`tenant\`, \`principal\`, and \`user\` — no consumer still reaches into the request context for app-lifetime services.

Closes INTR-66

### DIFF
--- a/packages/hub/src/app.test.ts
+++ b/packages/hub/src/app.test.ts
@@ -1,8 +1,11 @@
 import { describe, test, expect } from "bun:test";
 import { type } from "arktype";
+import { Hono } from "hono";
 import type { DB } from "@interchange/db";
-import { createApp } from "./app";
+import { createApp, createHubContextMiddleware, mountHubRoutes } from "./app";
+import type { AppEnv } from "./context";
 import { createEventCollectorRegistry } from "./event-collector-registry";
+import type { GetSession } from "./session";
 import type { SessionService } from "./session-service";
 import { createSidecarRouter } from "./ws/sidecar-handler";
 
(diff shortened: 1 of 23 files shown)


alexanderguy:
## Code Review

**No outstanding issues. Ready to merge.**

What I verified:

- Commit messages match diffs across all three commits.
- `c.get(...)` across `packages/hub/src` returns only the four per-request keys (`tenant`, `principal`, `user`, `session`); no consumer still reaches for a session-layer service through the request context.
- `packages/hub/src/app.test.ts:120` covers the new composition boundary: third-party `Hono<AppEnv>` + `createHubContextMiddleware` + `mountHubRoutes` + sibling route, all sharing the same `c.get("user")`. Second test confirms `mountHubRoutes` does not claim `/api/auth/*`.
- Grant enforcement is preserved: `authorize(...)` argument list in `packages/hub/src/middleware/grant.ts` is unchanged. `routes/instances.test.ts` still exercises "missing grant returns 403".
- `make all` green.
- No `.md` file references the deleted setter middleware, the request-context-bag shape, or the closed-Hono-constant pattern.

Design choices worth flagging (not findings, just things a reviewer might ask about):

- `packages/hub/src/app.ts:80-87` uses `??` to default `grantStore` / `conditionRegistry`. Intentional — the JSDoc above the function declares the defaults as the hub's standard choices.
- `packages/hub/src/middleware/grant.ts` does not export `ResourceFn`. The JSDoc on `RequireGrant` points at `idResource(...)` as the supported way to build the function form, which keeps the public surface narrow.
- `Handler` from hono is not re-exported from `@interchange/hub`. Composers using `mountHubRoutes` are already on hono and can import it directly.
- `createHubContextMiddleware` is currently a one-line delegate to `createSessionMiddleware`. It's exported as a factory because the Linear acceptance criteria call for one; it's the seam to grow into.

PR #24, opened 2026-05-19
Split the hub package into hub-api, hub-sessions, and hub-common

alexanderguy:
## Summary

- Splits `@interchange/hub` into three packages along the boundaries the internal refactor left behind: `@interchange/hub-api` for the HTTP surface (Hono app factory + context middleware + route group factories + middleware + better-auth wrapper + request/response helpers + request context + session contract + timeline read model), `@interchange/hub-sessions` for the session plane (sidecar wire layer + typed event emitter + session service + event collector and registry + agent repository + hub session orchestrator + credential-push pipeline), and `@interchange/hub-common` for genuinely shared hub utilities (today: the id generator and prefix table).
- Deletes the aggregated `@interchange/hub` package with no compatibility shims. `apps/hub`, `bin/gen-api-docs`, and the sidecar tests import from the new packages directly.
- Updates `LAYOUT.md` to reflect the realized package boundaries.

Closes INTR-67.

## Deviations from the issue body

- **Commit order swapped.** The issue listed hub-api as commit 1 and hub-sessions as commit 2. hub-api depends on hub-sessions, so creating hub-api first would have broken the build-every-commit invariant enforced by `make all`. Greybeard concurred with the swap before any code was written.
- **Five commits instead of four.** `packages/hub/src/ids.ts` (`generateId`) was used by both the api and the session halves. Introducing it as a new `@interchange/hub-common` package up front (commit 1) keeps the dependency direction one-way (`hub-api -> hub-sessions -> hub-common`) and removes the temptation to special-case ids into one half. The package README charters its scope so the generic name does not invite drift.
- **Placeholder removal is a no-op.** The issue body asked to drop deferred `hub-credentials` and `hub-ca` placeholders from `LAYOUT.md`. Neither existed on main, so that half is silent. The old `@interchange/hub` row mentioned "certificate authority" as a responsibility; that wording was dropped from the new rows because no such code exists in the repository.

## Per-commit shape

1. `Move generateId into a new @interchange/hub-common package` -- new package + 14 internal call-site rewires + drop of the unused `export { generateId }` re-export from the hub barrel.
2. `Extract the session plane into @interchange/hub-sessions` -- 17 file moves (R100) + new package barrel + 5 internal hub consumer rewires. The transitional `packages/hub` barrel re-exports the new surface so external consumers stay valid.
3. `Extract the HTTP surface into @interchange/hub-api` -- 32 file moves (R100) + new package barrel. `packages/hub` becomes a pure re-export of both new packages.
4. `Delete the aggregated @interchange/hub package` -- removes the package and rewires `apps/hub`, `bin/gen-api-docs`, `tests/sidecar/deploy-flow.test.ts`, `apps/sidecar/src/ws-client.test.ts`. OpenAPI document regenerates byte-identically because the route handlers did not move on disk, only the workspace they are exported from.
5. `Update LAYOUT.md for the hub package split` -- Control Plane table now lists hub-api, hub-sessions, hub-common; intro and trailing sentence updated.

## Test plan

- [x] `make all` (lint, prettier, `bin/gen-api-docs.ts --check`, `tsc -b --noEmit`, full `bun test` + sidecar/inference auxiliary suite) green at HEAD and at every intermediate commit
- [x] 1487 + 9 tests pass with zero failures
- [x] `docs/API.md` regenerates byte-identically
- [x] No `@interchange/hub` references remain in source, configs, or scripts


alexanderguy:
## Code review (`code-review` skill)

Reviewed the diff at HEAD against `origin/main`.

### Architecture

The split lands on a clean one-way dependency: `hub-api -> hub-sessions -> hub-common`. Nothing under `packages/hub-sessions/src/` reaches into hub-api, and the only direction of coupling between them is hub-api importing the session-service / sidecar-router / event-collector / credential-push surface that the routes need to operate on. This is the right cut.

`hub-common` exists because `generateId` is needed by both halves and is hub-schema-owner knowledge (the prefix table enumerates the hub's entity namespace). Forcing it into either side would have created an artificial dependency edge. The README charters the package's scope with four explicit gates against junk-drawer drift -- good defense for a generically-named package.

### Package shape

- `packages/hub-api/package.json:12-27` declares deps that match the moved files' actual imports. `@hono/standard-validator` is listed but not imported anywhere in the source -- this carried over from the old hub package.json and is pre-existing dead config; worth a follow-up cleanup but not blocking.
- `packages/hub-sessions/package.json` and `packages/hub-common/package.json` are minimal and accurate.
- All three new `tsconfig.json` files match the existing repo shape (`extends ../../tsconfig.base.json`, `include ["src/**/*.ts"]`); the root `tsconfig.json` adds the three references in alphabetical adjacency.
- Barrels (`packages/hub-api/src/index.ts`, `packages/hub-sessions/src/index.ts`) export everything consumers need through the named package; no deep imports survive in `apps/hub`, `bin/`, or `tests/`.

### Consumer rewires

`apps/hub/src/index.ts` splits cleanly: `createApp`/`createAuth` from hub-api, the seven session-plane symbols + `WsHandle` from hub-sessions. The previous inline `import("@interchange/hub").WsHandle` type annotation is replaced with a regular `WsHandle` reference, which reads better and confirms `WsHandle` is on the hub-sessions barrel rather than a deep import.

`bin/gen-api-docs.ts`, `tests/sidecar/deploy-flow.test.ts`, and `apps/sidecar/src/ws-client.test.ts` are single-line/single-block import switches. No deep imports.

`apps/hub/package.json` swaps the `@interchange/hub` dep for `@interchange/hub-api` + `@interchange/hub-sessions`. `@interchange/hub-common` is correctly omitted (it's transitive via hub-sessions).

### LAYOUT.md

The Control Plane table reflects the realized boundaries: three rows for hub-api, hub-sessions, hub-common, plus the unchanged hub-client and db rows. Each row describes what the package actually owns today. The "certificate authority" wording carried in the old hub row is dropped because no such code exists in the repository.

### Build / test

`make all` passes -- lint, prettier, `bin/gen-api-docs.ts --check` (confirms `docs/API.md` regenerates byte-identically), `tsc -b --noEmit`, full `bun test` (1487 pass), and the sidecar/inference auxiliary suite (9 pass). No test coverage was lost in the move -- every test file followed its source file.

### Pre-existing conventions preserved (out of scope, flagging for awareness)

- `drizzle-orm`, `isomorphic-git`, `openapi-types` are imported directly from `packages/hub-{api,sessions}/src/**` without being listed as direct deps -- they come transitively through `@interchange/db` and `@interchange/storage-isogit`. This mirrors the previous hub package.json.
- `apps/sidecar/package.json` does not list `@interchange/hub-sessions` despite `apps/sidecar/src/ws-client.test.ts` importing from it. Pre-existing pattern relying on workspace-level resolution.
- `@hono/standard-validator` listed but unused (mentioned above).

### Verdict

Clean. No blockers, no required changes. The three pre-existing-convention items above are worth a follow-up but are explicitly out of scope here.

PR #26, opened 2026-05-19
Drop opsh as a contributor prerequisite

alexanderguy:
## Summary

- Vendored opsh bumped to 0.9.0, which supports a source-mode dispatch (`BASH_SOURCE[0] != $0` exposes `SCRIPTDIR` without re-executing).
- Every `#!/usr/bin/env opsh` script in `.githooks/` and `bin/` switches to `#!/usr/bin/env bash` and sources the vendored `bin/opsh` as its first line, so opsh no longer needs to be on PATH.
- `bin/check-env` additionally verifies the vendored copy exists before sourcing it, so the env-validation gate fails with a clear diagnostic rather than a bash "no such file" error.
- README and `DEV.md` prerequisite lists drop the opsh entry; `DEV.md`'s "Bin Scripts" intro is rewritten to describe the new convention.

## Scope note

The issue called out three hooks plus `bin/check-env`. The wider scope here also migrates `bin/hub`, `bin/db-migrate`, `bin/db-reset`, and `bin/add-package` — confirmed with the issue author before implementing. Without those four, the README's quick-start (`bin/db-reset && bun bin/dev.ts --seed`) would still require opsh on PATH even though the prerequisite list no longer says so.

## Changes

- `bin/opsh`: replaced wholesale with the upstream 0.9.0 single-file release (GPL-3.0, attribution in the commit message).
- `.githooks/{commit-msg,post-checkout,pre-commit}`, `bin/{check-env,hub,db-migrate,db-reset,add-package}`: shebang -> `bash`, source the vendored opsh, `opsh::version::require` bumped to `v0.9.0`.
- `bin/check-env`: explicit pre-source existence check for `bin/opsh`, with a comment explaining why this script is the only one with the guard.
- `README.md`, `DEV.md`: drop opsh prerequisite; rewrite the bin-scripts intro.

## Testing

- `make all` succeeded with opsh deliberately stripped from PATH (verified via `env -i PATH=<no-opsh-dir> bash -c 'make all'`).
- `git commit` and `git commit --allow-empty` both succeeded under the same stripped PATH; the `commit-msg` and `pre-commit` hooks fired and validated under the new mechanism.
- The three real commits on this branch passed both hooks during their own commit, exercising the new code paths end-to-end.

## Known latent issue (not addressed)

`bin/opsh:411` -- `ssh::begin` writes a proxy `ssh` shim with `#!/usr/bin/env opsh`. The path is unreachable in this repo (no caller of `ssh::begin`/`ssh::end`), so it does not regress the goal of this PR. Vendored unmodified per attribution discipline; worth a follow-up upstream.

Closes INTR-71

### DIFF
--- a/.githooks/commit-msg
+++ b/.githooks/commit-msg
@@ -1,6 +1,7 @@
-#!/usr/bin/env opsh
+#!/usr/bin/env bash
 # shellcheck shell=bash
 
+source "$(dirname -- "${BASH_SOURCE[0]}")/../bin/opsh"
 source "$SCRIPTDIR/lib.opsh"
 
 MSGFILE=$1
--- a/.githooks/post-checkout
+++ b/.githooks/post-checkout
@@ -1,5 +1,7 @@
-#!/usr/bin/env opsh
+#!/usr/bin/env bash
 # shellcheck shell=bash
+
+source "$(dirname -- "${BASH_SOURCE[0]}")/../bin/opsh"
 
 PREV_HEAD=$1
 shift
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,6 +1,7 @@
-#!/usr/bin/env opsh
+#!/usr/bin/env bash
 # shellcheck shell=bash
 
+source "$(dirname -- "${BASH_SOURCE[0]}")/../bin/opsh"
 source "$SCRIPTDIR/lib.opsh"
 
 step::000::checkout-staging() {
--- a/DEV.md
+++ b/DEV.md
@@ -3,7 +3,6 @@
 ## Prerequisites
 
 - [Bun](https://bun.sh/) (1.2+)
-- [opsh](https://github.com/alexanderguy/opsh) (v0.7+) -- for the shell scripts in `bin/`
 - PostgreSQL (15+)
 - Git hooks configured: `git config core.hooksPath .githooks`
 
@@ -135,7 +134,7 @@ is validated.
 
 ## Bin Scripts
 
-All scripts live in `bin/`. Shell scripts that start with `#!/usr/bin/env opsh` use the bundled `opsh` shell framework (`bin/opsh`).
+All scripts live in `bin/`. The bash scripts source the bundled [opsh](https://github.com/alexanderguy/opsh) framework as a library from `bin/opsh`, so opsh does not need to be installed separately.
 
 | Script                | Usage                                            | Description                                                                                             |
 | --------------------- | ------------------------------------------------ | ------------------------------------------------------------------------------------------------------- |
--- a/README.md
+++ b/README.md
@@ -38,11 +38,10 @@ bin/db-reset && bun bin/dev.ts --seed
 Starts the hub, sidecar, and UI — entry points in
 [`apps/`](./apps).
 
-Requires [Bun](https://bun.sh/) 1.2+,
-[opsh](https://github.com/alexanderguy/opsh) 0.7+, and PostgreSQL
-15+. See [`DEV.md`](./DEV.md) for everything else — environment
-files, role setup, default ports, seed credentials, partial-stack
-variants, reset recipes.
+Requires [Bun](https://bun.sh/) 1.2+ and PostgreSQL 15+. See
+[`DEV.md`](./DEV.md) for everything else — environment files, role
+setup, default ports, seed credentials, partial-stack variants,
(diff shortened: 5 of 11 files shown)


alexanderguy:

Dispatched the `critique` subagent to diff the branch against `origin/main` and apply the `code-review` checklist. Full punch list below, followed by disposition.

### Findings

**Correctness / safety**

- `bin/opsh:411` (vendored 0.9.0) -- bundled `ssh::begin` writes a proxy `ssh` shim with `#!/usr/bin/env opsh`. Pure latent issue (no in-repo caller of `ssh::begin`/`ssh::end`), so it does not regress this PR's goal today. The prerequisite list now says "no opsh needed" while a code path inside the vendored framework would still require opsh on PATH at runtime. Note for follow-up; not a push blocker since opsh is vendored unmodified and the path is unreachable from this repo.

- `bin/check-env:4-13` -- pre-source guard checks `[[ ! -f $OPSH_VENDORED ]]` and fails with a hand-rolled `echo ... >&2; exit 1`. Asymmetric with the other migrated scripts, which `source` the vendored opsh with no pre-check. The reviewer flagged this as either drop-or-extend-uniformly; also noted the original commit-message wording "missing or corrupt" overstated what the guard catches (only "missing"; a corrupt opsh still fails generically via bash).

- Migrated scripts use `source "$(dirname -- "${BASH_SOURCE[0]}")/..."` without resolving symlinks. Pre-existing behavior pattern (old `#!/usr/bin/env opsh` had the same fragility via PATH lookup), and the repo has no symlinks pointing into `bin/`. Not a push blocker.

**Scope**

- Migration is complete across all eight in-scope shell scripts. The remaining `bin/` shell scripts (`bin/audit`, `bin/posix-demo`, `bin/ring-demo`) are plain `#!/usr/bin/env bash` with no opsh dependency, verified.
- `git grep -n 'opsh'` across `.md`/`Makefile`/`*.json`/`*.ts` yields only the single intentional reference in `DEV.md:137`. No stray prerequisite mentions left behind.

**Build / test**

- Reviewer independently re-verified the load-bearing cases: (a) `env -i PATH=<minimal>:... bash -c './bin/check-env'` from repo root passes all four steps; (b) `env -i ... bash -c 'git commit --allow-empty ...'` fires `pre-commit` + `commit-msg` cleanly, runs `make lint` inside the staging tree, and commits successfully; (c) vendored `bin/opsh` md5 (`03dc4604dc7097553e7b0f94c9880f8f`) matches the upstream 0.9.0 release asset byte-for-byte.

**Convention / commit coherence**

- 463d14a "Vendor opsh 0.9.0": diff is 100% `bin/opsh`. Attribution block (source URL, asset, version, license, author, retrieval date) present per the `style` skill's external-code rule. No mixed changes.
- 6fdf69e "Source the vendored opsh instead of relying on the kernel": diff is the 8 script migrations. Message's "exactly matching" claim about SCRIPTDIR is accurate for all in-repo callers; small absolute-vs-relative drift in source mode is benign for current call sites.
- dac3d71 "Stop listing opsh as a contributor prerequisite": diff is exactly the two doc files. Message matches diff.
- Commits read as a story: vendor -> switch call sites -> drop the prerequisite. No interleaved refactor/feature.
- `opsh::version::require` bumped from `v0.7.0` -> `v0.9.0` in the five `bin/` scripts. The three `.githooks/*` scripts still do not call `opsh::version::require`; pre-existing structure, not a regression.

### Disposition

- `bin/opsh:411` `ssh::begin` proxy shim: **deferred** -- vendored unmodified per attribution rule; path unreachable in this repo. Surfaced above for follow-up upstream.
- `bin/check-env` asymmetric pre-source guard: **fixed in 6fdf69e (post-review fixup folded in via autosquash)** -- added a comment to `bin/check-env` explaining that this script is the env-validation gate, which is why it has the guard and the other seven migrated scripts do not.
- Commit-message "missing or corrupt" inaccuracy: **fixed in 6fdf69e (reword in the same rebase)** -- now reads "a missing bin/opsh fails with a clear diagnostic rather than bash's generic 'no such file' error", matching what the guard actually catches.
- Symlink fragility: **noted** -- pre-existing pattern, not a regression.
- SCRIPTDIR absolute-vs-relative drift: **noted** -- works for all current call sites; heads-up only.
- Hooks don't `opsh::version::require`: **out of scope** -- pre-existing; hooks ship with `bin/opsh` so version drift is unlikely.

PR #28, opened 2026-05-21
Add per-call inactivity and total timeouts to inference harness

alexanderguy:
## Summary

- Inference harness arms an inactivity timer (reset on every wire chunk) and a total wall-clock cap per call; either firing aborts the fetch and produces an `inference.error` with the new `"timeout"` category naming which timer fired and its threshold.
- Adds a `scenario.stall()` helper to `@intx/inference-testing` with a `StallHandle` exposing `aborted` + `awaitAbort` telemetry, so timeout tests can assert directly on `AbortController` propagation rather than only on the downstream error event.
- Timer + signal-listener lifecycle is owned by a single `try/finally` around the generator body, so cleanup runs on every exit path including consumer abandonment, non-OK HTTP responses, and null-body responses.
- Error response bodies are now read once as text and then parsed (fixing a pre-existing WHATWG locked-body bug where the `.json()` → `.text()` fallback always threw), and plain-text bodies are surfaced as the `InferenceError.message` (bounded at 500 chars; full body retained in `error.raw`) so user-facing replies, timeline parts, and audit store rows carry actionable diagnostic content instead of a generic statusText echo.

## Changes

- `packages/types/src/runtime.ts`: `"timeout"` added to `InferenceError` category union; `inactivityTimeoutMs?` / `totalTimeoutMs?` added to `InferenceOptions` with 120k / 600k defaults.
- `packages/inference/src/errors.ts`: `classifyTimeoutError(kind, thresholdMs)` helper.
- `packages/inference/src/harness.ts`: two timers, `AbortController`, `combineSignals` returning `{ signal, cleanup }`, signal-bound `awaitWithSignal` for non-OK error-body reads, text-then-parse pattern for body extraction, plain-text fallback with truncation in `extractErrorMessage`, generator body wrapped in `try/finally` with single cleanup owner.
- `packages/inference/src/index.ts`: exports `Scheduler` + `createDefaultScheduler`.
- `packages/inference-testing/src/harness.ts`: `enableInferenceTimers` opt-in wires `deps.scheduler` to the virtual clock; default is an inert scheduler so other test suites are unaffected.
- `packages/inference-testing/src/scenario.ts`: `scenario.stall({ predicate? })` returning `StallHandle`; abort observer wired into `routeWaitingFetch`; `dispose()` resolves any pending `awaitAbort` promises.
- `tests/inference/timeout.test.ts`: six scenarios — inactivity fires on stall, slow-but-steady stream survives, total cap fires on active stream, healthy short call passes under defaults, AbortController fires on timeout, per-call override on a shared one-second-gap wire.
- `tests/inference/lifecycle.test.ts`: four regression tests across three describe blocks — non-streaming exit paths (non-OK HTTP, null body), consumer abandonment, caller-signal listener accounting.
- `tests/inference/error-body.test.ts`: four cases covering structured JSON envelope extraction, plain-text body passthrough, long-text truncation with marker + full `error.raw`, and empty-body `statusText` fallback.
- `docs/INFERENCE.md`: `"Timeout"` entry in Error Categories; new `Per-Call Timeouts` section with defaults, override mechanism, reasoning-model tuning guidance, and a clarification that the inactivity timer resets on every wire chunk rather than strictly on yielded events.
- `packages/inference-testing/README.md`: `Stall scenarios` subsection documenting `scenario.stall` and `StallHandle`.

## Testing

- [x] `make all` green: lint + `tsc -b --noEmit` + 1502 main tests + 9 sequential tests passing.
- [x] New timeout-behaviour tests in `tests/inference/timeout.test.ts` — six scenarios cover the issue's "tests first" checklist.
- [x] New lifecycle regression tests in `tests/inference/lifecycle.test.ts` — four invariants pinned across three describe blocks (timer cancellation on every exit path; no caller-signal listener accumulation).
- [x] New error-body extraction tests in `tests/inference/error-body.test.ts` — four cases pinning the public-facing shape of `InferenceError.message` and `error.raw` across the input shapes.
- [x] Existing `@intx/inference` and `@intx/inference-testing` suites continue to pass without modification.
- [x] Self-review via the `code-review` skill: four rounds. Round-one surfaced six findings (timer leaks, signal-listener leak, no-finally on generator, unbounded error-body read, two doc errors); all six resolved. Round-two clean. Round-three covered the body-read fix and a rebase that landed two doc/test refinements via `edit` (not fixups) so they live in the right commits — clean, with a single non-blocking observation (plain-text bodies captured in `error.raw` but not in `error.message`). Round-four covered the plain-text-message follow-up that closed that observation — clean.

Closes INTR-87

### DIFF
--- a/docs/INFERENCE.md
+++ b/docs/INFERENCE.md
@@ -656,12 +656,28 @@ Provider errors are classified into categories that determine the reactor's resp
 - **Quota exhausted** — Provider-level usage limit hit (distinct from transient rate limits). Response: fail or switch model, depending on director policy.
 - **Fatal** — Invalid request, unsupported model, malformed content. Response: fail immediately with diagnostic information.
 - **Aborted** — Caller cancelled via AbortSignal. Response: clean termination.
+- **Timeout** — Per-call inactivity or total wall-clock cap fired (see Per-Call Timeouts). The call produced no usable response. Response: treat as transient infrastructure failure and retry per director policy rather than as a model decision.
 
 The classifier inspects HTTP status codes, error response bodies, and provider-specific error message patterns. The classified error is delivered to the director as an `inference.error` event, and the director decides the response — retry, compact, switch model, suspend, or fail. The error categories inform the director's decision but do not hardcode the response.
 
 ### Retry Behavior
 
 Retryable errors use exponential backoff: `baseDelay * 2^attempt`. The director controls maximum attempts and total budget. Failed attempts are removed from the message history before retrying — the model should not see its own error responses.
+
+## Per-Call Timeouts
+
+Every `runInference` call arms two timers: an inactivity timer reset on every chunk the harness receives from the wire, and a total wall-clock cap that starts at `fetch()`. When either timer fires, the harness aborts the underlying fetch's combined `AbortController` and emits an `inference.error` with category `"timeout"` and a message naming which timer fired and its threshold value. The combined controller wires together the caller-supplied signal (if any) and the harness's internal timeout signal, so a fetch implementation that respects `AbortSignal` sees both.
+
+"Wire chunk" means any SSE data line the harness reads, not strictly the events the harness yields downstream — heartbeat frames, sentinel-only chunks (e.g. `[DONE]`), and trailing usage events all count as activity. This is the more permissive of the two natural rules: it treats any sign of provider liveness as defeat of the inactivity timer, which is what "stalled SSE stream" intuitively means and what the safety property actually guards against.
+
+The defaults are conservative:
+
+- `inactivityTimeoutMs: 120_000` (2 min). Reasoning-heavy models emit tokens regularly when actually thinking; two minutes of pure silence is real stall, not legitimate reasoning. Tune higher for models with extended pure-silence thinking stretches — the `interchange-demo-dispatch` planner agent was observed reasoning for roughly seven minutes between tool calls during early benchmarking, which would require a higher inactivity threshold.
+- `totalTimeoutMs: 600_000` (10 min). Matches Anthropic's documented per-call recommendation and fits within typical CI budgets. Backstop for streams that keep emitting forever without terminating.
+
+Both can be overridden per call via `InferenceOptions.inactivityTimeoutMs` and `InferenceOptions.totalTimeoutMs`. Set to a smaller number to fail fast; set to a larger number to accommodate slower model behaviour.
+
+A timeout always classifies as `"timeout"`, never as `"aborted"` — even though the underlying mechanism is an `AbortController` firing. Distinguishing the two at the category level lets directors apply different policies: a caller-initiated abort is the user's decision and should not be retried, while a timeout is an infrastructure failure and may be retried.
 
 ## Abort Handling
 
--- a/packages/inference-testing/README.md
+++ b/packages/inference-testing/README.md
@@ -165,6 +165,45 @@ harness.scenario.abortAfter(
 stream and errors its controller — useful for modelling network failures
 mid-response.
 
+### Stall scenarios
+
+`scenario.stall` wires up a parked-stream fixture for tests that exercise
+the inference layer's per-call timeouts. The matched fetch is bound to a
+stream the helper never feeds, so the SSE iterator parks on its first
+read; in virtual time the test advances past the timeout threshold and
+the inactivity (or total) timer in `runInference` aborts the underlying
+fetch. The returned `StallHandle` exposes live abort telemetry so tests
+can assert on `AbortController` propagation directly, not only on the
+downstream `inference.error` event the abort produces.
+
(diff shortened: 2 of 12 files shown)


alexanderguy:

9 commits (`cfbe6f8..a991ba3`) introducing per-call inactivity + total timeouts to the inference harness, with adjacent error-body handling improvements that emerged during review.

### What the code does

**Inference timeouts.** Every `runInference` call arms two timers: an inactivity timer reset on each wire chunk and a total wall-clock cap from fetch onward. Either firing aborts a combined `AbortController` (caller signal + internal timeout signal) and emits an `inference.error` with the new `"timeout"` category, a message naming which timer fired and its threshold. Defaults are 120 s / 600 s, overridable per call via `InferenceOptions.inactivityTimeoutMs` / `totalTimeoutMs`. `combineSignals` returns `{ signal, cleanup }` so the abort listeners it installs on the caller signal are explicitly removed on every exit path — no listener accumulation against long-lived caller signals. A `Scheduler` abstraction backs the timers; injectable via `Dependencies.scheduler` for deterministic testing, defaults to a `setTimeout`-wrapping production scheduler.

**Lifecycle hygiene.** The generator body is wrapped in `try/finally` with a single cleanup owner (`cleanupTimers()` + `cleanupSignal()`). The finally runs on normal completion, early `return`, thrown errors, and consumer abandonment via `for await … break`. All inline cleanup calls inside the body were removed — exactly one cleanup site.

**Error-body extraction.** `!response.ok` reads the body as text once via `awaitWithSignal` (signal-bound to prevent hostile servers from hanging the call past the total-timeout horizon), then attempts `JSON.parse(text)`. Structured envelopes (`{ error: { message } }`, `{ message }`) extract their nested message field as before. Plain-text bodies now surface as the message, bounded at 500 chars with a `… (truncated; full body in error.raw)` marker. `error.raw` retains the untruncated body for audit-time inspection.

**Test seam.** New `scenario.stall({ predicate? })` helper on `@intx/inference-testing` returns a `StallHandle` exposing the stalled stream, an `aborted` flag, and an `awaitAbort` promise that flip / resolve when the matched fetch's signal aborts. Test harness gains an `enableInferenceTimers` opt-in that backs the inference scheduler with the virtual clock so timeout tests fire at virtual time; default off so other test suites are unaffected.

**Tests.** 14 new tests across three files:
- `tests/inference/timeout.test.ts` (6): inactivity fires on stall, slow-but-steady survives, total cap fires on active stream, healthy short call passes under defaults, `AbortController` fires on timeout, per-call override on a shared one-second-gap wire.
- `tests/inference/lifecycle.test.ts` (4 in 3 describe blocks): non-streaming exits (non-OK HTTP, null body), consumer abandonment via `break`, caller-signal listener accounting.
- `tests/inference/error-body.test.ts` (4): JSON envelope extraction, plain-text passthrough, long-text truncation, empty-body statusText fallback.

**Documentation.** `docs/INFERENCE.md` gains a `"Timeout"` entry in Error Categories and a `Per-Call Timeouts` section covering defaults, the override mechanism, reasoning-model tuning guidance, and a clarification that the inactivity timer resets on every wire chunk (heartbeats, sentinel-only frames, trailing usage) rather than only on yielded events. `packages/inference-testing/README.md` gains a `Stall scenarios` subsection documenting `scenario.stall` and `StallHandle`.

### Code-quality assessment

- **Conventions.** `camelCase` functions, `PascalCase` types, acronym casing preserved (`SSE`, `HTTP`, `JSON`, `URL`). `import type` under `verbatimModuleSyntax`. Runtime types declared via arktype with derived TypeScript types alongside. No `as` assertions in production code; the single test-file assertion has an `eslint-disable-next-line` with a justification per `CONVENTIONS.md`. No console, no default exports, no emojis.
- **Defensive coding.** No silent error swallowing introduced; all error paths emit typed `inference.error` events. The `combineSignals` `cleanedUp` flag makes cleanup idempotent. `awaitWithSignal` removes its own listener on every settlement path. The scheduler abstraction's cancellers are documented idempotent.
- **Scope.** No drive-by changes. Each commit's diff matches its message. The body-read fix and the plain-text-surfacing fix are correctly scoped as separate commits from the timeout machinery — different concerns with different rationales.
- **Backwards compatibility.** `InferenceError.category` adds `"timeout"` (additive). `InferenceOptions` adds two optional fields (additive). `Dependencies.scheduler` is optional with a sensible default. Existing callers continue to work unchanged.

### Known limitation, non-blocking

`extractErrorMessage` applies the 500-char cap only to the new plain-text fallback branch. Structured JSON envelopes are unchanged — a misbehaving server returning `{ "message": "<200KB>" }` would still produce an unbounded `error.message`. Practical risk is low (server-curated message fields are concise in practice) but worth a Linear ticket if it ever bites in production. One-call-site change to close the gap.

### Verdict

Ready to merge.

PR #29, opened 2026-05-21
Throw ProtocolMismatchError instead of silently dropping malformed SSE chunks

alexanderguy:
## Summary

- OpenAI and Anthropic SSE adapters now throw `ProtocolMismatchError` when `JSON.parse` fails or an arktype validator rejects the chunk, instead of logging a warning and returning `[]`. The harness's existing stream-error catch maps that throw to an `inference.error` event with the new `"protocol_mismatch"` category, carrying the offending bytes or parsed object in `error.raw` and the JSON parse error or schema summary in `error.message`.
- Closes the Fireworks symptom that motivated INTR-84: `OpenAIToolCallDelta` now accepts `string | null` on `id`, `function.name`, and `function.arguments`, with `null → undefined` normalisation at the consumer site. Start and fragment branches are independent so a single delta can emit both a `tool_call.start` and a `tool_call.delta` (which is what Fireworks does on the first fragment delta).
- The pre-existing `input_json_delta`-without-`tool_use_start` throw in the Anthropic adapter (previously a bare `Error` classified as `retryable`) becomes a `ProtocolMismatchError` too — out-of-order events are a protocol violation, not a transport flake.

## Changes

- `packages/types/src/runtime.ts`: `"protocol_mismatch"` added to `InferenceError.category` union.
- `packages/inference/src/adapter.ts`: `ResponseParser` contract docstring amended to permit `ProtocolMismatchError` throws and explicitly note that adapter-emitted `inference.error` / `inference.done` events are silently dropped by the harness.
- `packages/inference/src/errors.ts`: new `ProtocolMismatchError` class with a `raw` field; new `classifyProtocolMismatch(detail, raw?)` helper; `classifyStreamError` dispatches `ProtocolMismatchError` to the new category.
- `packages/inference/src/index.ts`: exports `ProtocolMismatchError` and `classifyProtocolMismatch`.
- `packages/inference/src/providers/openai.ts`: schema relaxation (Fireworks-shape `null`-on-inner-fields), `null → undefined` normalisation, independent start and fragment emission branches, `ProtocolMismatchError` throws on JSON parse and schema rejection paths. Now-unused LogTape import dropped.
- `packages/inference/src/providers/anthropic.ts`: `ProtocolMismatchError` throws on JSON parse, schema rejection, AND the `input_json_delta`-without-preceding-`tool_use_start` path. Now-unused LogTape import dropped.
- `docs/INFERENCE.md`: new `"Protocol mismatch"` entry in the Error Categories list.

## Testing

- [x] `make all` green: lint + `tsc -b --noEmit` + 1514 main tests + 9 sequential tests passing.
- [x] Fireworks regression test in `tests/inference/providers/openai.test.ts` — hand-rolls the three-chunk sequence (start delta with `name`, two fragment deltas with `id: null` / `name: null`) and asserts the start event plus accumulated argument fragments.
- [x] Dual-emission test in the same file — pins the consumer-site claim that one chunk can carry both a start signal and a non-empty argument fragment.
- [x] Adapter-level unit tests for `ProtocolMismatchError` throws across both providers (JSON parse path, schema rejection path, and for Anthropic the `input_json_delta`-without-tool_use_start path).
- [x] End-to-end test (`tests/inference/protocol-mismatch.test.ts`) drives `runInference` against a stub fetch returning malformed SSE bytes and asserts the emitted `inference.error` has the right category, message, and raw payload for both providers.
- [x] `classifyStreamError` unit tests cover the new `ProtocolMismatchError → protocol_mismatch` dispatch.
- [x] Self-review via the `code-review` skill: three rounds (initial via critique subagent, follow-up after fixup-and-rebase pass, final after a second rebase-with-edits pass). Final pass clean.

## Out of scope

The third checklist item on INTR-84 (logger silent-init defensive observability) is **deferred to INTR-91** by user decision. The `@intx/inference` adapters no longer rely on the logger to surface schema-rejection diagnostics — those flow through typed `inference.error` events instead — so the logger fallback is now genuinely secondary.

Closes INTR-84

### DIFF
--- a/docs/INFERENCE.md
+++ b/docs/INFERENCE.md
@@ -657,6 +657,7 @@ Provider errors are classified into categories that determine the reactor's resp
 - **Fatal** — Invalid request, unsupported model, malformed content. Response: fail immediately with diagnostic information.
 - **Aborted** — Caller cancelled via AbortSignal. Response: clean termination.
 - **Timeout** — Per-call inactivity or total wall-clock cap fired (see Per-Call Timeouts). The call produced no usable response. Response: treat as transient infrastructure failure and retry per director policy rather than as a model decision.
+- **Protocol mismatch** — Upstream emitted a chunk that violates the provider's streaming protocol: malformed JSON, schema validation failure, or an out-of-order event sequence; the offending chunk is preserved in `error.raw`. Response: fail rather than retry — retrying the same request won't fix the bytes.
 
 The classifier inspects HTTP status codes, error response bodies, and provider-specific error message patterns. The classified error is delivered to the director as an `inference.error` event, and the director decides the response — retry, compact, switch model, suspend, or fail. The error categories inform the director's decision but do not hardcode the response.
 
--- a/packages/inference/src/adapter.ts
+++ b/packages/inference/src/adapter.ts
@@ -26,8 +26,15 @@ export type RequestBuilder = (
 // requests.
 //
 // The parser may return an empty array for events it doesn't care about
-// (e.g., Anthropic's ping events). It must not throw; errors should be
-// returned as inference.error events.
+// (e.g., Anthropic's ping events). It MAY throw `ProtocolMismatchError`
+// when the upstream chunk violates the provider's protocol (malformed
+// JSON, schema validation failure, out-of-order events); the harness's
+// stream-error catch converts that into an `inference.error` with
+// category `"protocol_mismatch"` via `classifyStreamError`. No other
+// throw type is permitted, and adapter-returned `inference.error` or
+// `inference.done` events are silently dropped — the harness owns
+// emission of those terminal types and the only path through which
+// adapter-detected failures can surface is `ProtocolMismatchError`.
 export type ResponseParser = (sseData: string) => InferenceEvent[];
 
 // An adapter pairs a request builder with a response parser. Registration
--- a/packages/inference/src/errors.test.ts
+++ b/packages/inference/src/errors.test.ts
@@ -4,6 +4,8 @@ import {
   classifyNetworkError,
   classifyAbortError,
   classifyStreamError,
+  classifyProtocolMismatch,
+  ProtocolMismatchError,
 } from "./errors";
 
 describe("classifyHTTPError", () => {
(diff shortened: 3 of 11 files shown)


alexanderguy:

5 commits (`166652c..e4a3fb4`) replacing the silent-degrade pattern in both first-party SSE adapters with a typed throw that surfaces as a new `"protocol_mismatch"` `InferenceError` category. Originated as a Fireworks-shape `null`-on-tool-call-delta fix; broadened during planning to address the structural silent-degrade pattern that INTR-84 documented as the primary reason the original Fireworks bug took an afternoon to diagnose rather than fifteen minutes.

### What the code does

**Schema relaxation (`packages/inference/src/providers/openai.ts`).** `OpenAIToolCallDelta` accepts `string | null` on `id`, `function.name`, and `function.arguments`; the consumer site normalises `null → undefined` before the start/fragment branches. Fireworks emits these fields as literal `null` on every delta after the start; arktype's `"string"` rejected null and silently dropped every fragment, leaving `argsBuffer` empty and the finalised tool call with `arguments: {}`. The start and fragment branches are independent so a single delta can legitimately carry both signals (which is what Fireworks does on the first fragment delta).

**Protocol-mismatch category (`packages/types/src/runtime.ts`, `packages/inference/src/errors.ts`).** New `"protocol_mismatch"` member on the `InferenceError.category` union. New `ProtocolMismatchError` class carries a `raw: unknown` field alongside the standard `Error` fields. New `classifyProtocolMismatch(detail, raw?)` helper constructs the inference error and omits `raw` when undefined (per `exactOptionalPropertyTypes`). `classifyStreamError` dispatches `ProtocolMismatchError` to the protocol-mismatch classifier; everything else falls through to the existing `retryable` path. The `ResponseParser` adapter contract is amended to permit this one throw type and explicitly notes that adapter-emitted `inference.error` / `inference.done` events are silently dropped by the harness — `ProtocolMismatchError` is the only path through which adapter-detected failures surface.

**Adapter conversion (`packages/inference/src/providers/{openai,anthropic}.ts`).** Both adapters' `parseResponse` functions now throw `ProtocolMismatchError` on JSON parse failure and on arktype validation failure, instead of returning `[]`. The Anthropic adapter also converts its pre-existing `input_json_delta`-without-preceding-`tool_use_start` throw (previously a bare `Error` classified as `retryable`) to the same `ProtocolMismatchError` — out-of-order events are a protocol violation, not a transport flake. Both adapters drop their now-unused LogTape imports.

**Documentation (`docs/INFERENCE.md`).** New `"Protocol mismatch"` entry in the Error Categories section, matching the cadence of the existing bullets — trigger sentence + "Response:" guidance, with a parenthetical note that `error.raw` preserves the offending chunk.

**Tests.** 18 new test cases across four files. Adapter-level unit tests pin the throw shape and `raw` payload for both providers across all three throw sites (JSON parse, schema rejection, Anthropic-specific out-of-order). A Fireworks regression test hand-rolls the three-chunk sequence the issue documented (start delta with `name`, two fragment deltas with `id: null` / `name: null`) and asserts the start + accumulated fragments. A dual-emission test pins the consumer-site claim that one chunk can carry both signals. An end-to-end test (`tests/inference/protocol-mismatch.test.ts`) drives `runInference` against a stub fetch returning malformed SSE bytes and asserts the emitted event shape for both providers — the boundary every downstream consumer (default director's reply, hub event collector, audit store) actually reads.

### Code-quality assessment

- **Conventions.** New code uses `camelCase` for functions / `PascalCase` for types, preserves acronym casing (`SSE`, `HTTP`, `JSON`), uses `import type` under `verbatimModuleSyntax`, declares runtime types via arktype with derived TypeScript types alongside. No `as` assertions, no `console`, no default exports, no emojis. Adapter contract docstring uses "schema validation failure" wording rather than leaking arktype as the implementation detail.
- **Defensive coding.** No silent error swallowing introduced; the three throw sites in each adapter are loud. `ProtocolMismatchError.raw` always carries the offending data for operator inspection. `classifyProtocolMismatch` omits `raw` when undefined to satisfy `exactOptionalPropertyTypes`.
- **Scope.** Each commit's diff matches its message. The `input_json_delta` conversion in the Anthropic commit is explicitly documented in the commit body so a reader of `git log` isn't surprised by the diff. No drive-by changes in any commit.
- **Backwards compatibility.** `InferenceError.category` adds `"protocol_mismatch"` (additive, no break to arktype consumers). `ResponseParser` contract tightens — `parseResponse` may now throw `ProtocolMismatchError` — but no in-tree caller relied on the previous "must not throw" prohibition, so this is documentation-only at the call-site level.

### Deferred to follow-up

INTR-84's third checklist item — `@intx/log`'s `logger.warn` being a no-op when `setup()` was never called — is filed as INTR-91 by user decision. The motivation is mostly moot now: schema-rejection diagnostics flow through typed `inference.error` events (which downstream consumers always observe), so the logger fallback is operator-friendliness rather than correctness. INTR-91 closes the gap when prioritised.

### Verdict

Ready to merge.

PR #31, opened 2026-05-22
Introduce InferenceSource as the agent runtime's selection primitive

alexanderguy:
Closes INTR-90.

## Summary

- Replace `ProviderConfig` with `InferenceSource` end-to-end (types, wire frames, agent, harness, sidecar, hub, examples, tests). The new shape carries credentials, a stable routing `id`, and optional `defaults`/`capabilities`; per-call `InferenceOptions` merge over `source.defaults` once at the top of `runInference`. Pre-catalog the `id` is synthesized from `provider+model`; when the catalog lands the `id` becomes the catalog primary key.
- Drop the dead director-requested `model` parameter. With `source.model` canonical, the agent's wrapper-substitution layer disappears entirely.
- Refresh examples and demos to current Anthropic model IDs (retired 3.5 → `claude-sonnet-4-6` / `claude-haiku-4-5-20251001`; deprecated `claude-sonnet-4-20250514` → `claude-sonnet-4-6`).
- Move `bin/posix-demo.ts` and `bin/ring-demo.ts` into proper `examples/` workspace packages (now in the `tsc -b` graph and gated by `make all`).
- Capture and echo Anthropic thinking-block signatures. Pre-existing on `main`; multi-turn Anthropic conversations with extended thinking enabled were 400-ing without it. Landed here because the same Anthropic adapter is what the new model defaults route through.
- Consolidate per-credential resolution into a single helper consumed by both the launch route and the background source pushers.

## Commits

1. `069a348` Introduce InferenceSource as the agent runtime's selection primitive
2. `317dd43` Drop the director-requested model from the infer action
3. `4558c3c` Refresh example and demo model identifiers to current Anthropic IDs
4. `cee4758` Move the multi-agent demos from bin/ into examples/
5. `c1b0d83` Capture and echo Anthropic thinking block signatures
6. `b16d57c` Consolidate per-credential resolution into one helper

## Notes on related work

- INTR-79's description has been amended to reference commit `c1b0d83` as partial progress on the thinking-block work, and to spell out the remaining multi-block + redacted-thinking scope that the multimodal refactor needs to cover.
- A pre-existing `posix-demo` issue surfaced during validation: alpha's `connector.reply` events emit correctly but the user's INBOX watcher never fires, so the demo runs to its safety timer instead of the intended clean exit. The routing logic was untouched by this branch — same behavior reproduces on `main`. Worth filing a follow-up; not in scope here.

## Test plan

- [x] `make all` clean: 1554 tests pass (1545 main + 9 separated)
- [x] Live API smoke tests against Anthropic with the new model IDs:
  - `agent-quickstart` → `claude-sonnet-4-6` ✅
  - `agent-multi-provider` short prompt → `claude-haiku-4-5-20251001` (cheap tier) ✅
  - `agent-multi-provider` long prompt → `claude-sonnet-4-6` (smart tier) ✅
  - `coding-agent` → `claude-sonnet-4-6` ✅
- [x] Both moved demos execute end-to-end (`posix-demo`: alpha+beta exchange mail; `ring-demo`: prompt circulates the full ring alpha→bravo→charlie→delta→echo→alpha)
- [x] New focused tests cover the `source.defaults` merge precedence, the source registry rename (with duplicate-id rejection), and the thinking-block signature round-trip (parser, wire DSL, request builder)


alexanderguy:

Two rounds of `critique` during development plus a final pass post-rebase. Three minor findings; all three addressed on the branch.

### Major

None.

### Minor — all addressed

1. **Exhaustiveness guard for optional fields.** `applyInferenceSourceFields` in `packages/types/src/runtime.ts` used `{ readonly [K in keyof InferenceSource]: true }`, which TypeScript treats as also-optional for optional source keys. A future `region?: string` added to `InferenceSource` without being handled in the helper would silently fall through — the doc comment promises otherwise. Switched to `Required<InferenceSource>` so optional keys are also required in the guard (folded into `069a348`).

2. **Duplicate source ids silently drop the second entry.** Pre-catalog the `id` is `${provider}:${model}`, so two credential requirements resolving to the same plugin produce identical ids. `createSourceRegistry` in `packages/agent/src/source.ts` was using `find()` and silently picking the first — the second source became unreachable. Added a `Set<string>`-backed duplicate check that throws `InvalidInferenceSourceError` at registry construction, plus a test in `source.test.ts` (folded into `069a348`).

3. **Two implementations of "agent definition → InferenceSource[]".** `packages/hub-api/src/routes/instances.ts` and `packages/db/src/credential-resolution.ts` carried identical per-requirement loop logic, differing only in error handling. Extracted as `resolveOneCredential` returning a discriminated `CredentialOutcome`; each call site maps outcomes to its own convention (route → 409 codes, resolver → log+continue). Same operator-visible error messages preserved. Landed as `b16d57c`.

### Verified clean

- **Commit-message coherence**: each of six commits' diff matches its message after the rebase. No drive-by edits.
- **Pre-existing posix-demo prompt edits**: not present. The cli.ts diff vs `origin/main` shows only the rename, model refresh, and dead `case "message.received"` removal.
- **`noUncheckedIndexedAccess` guards in the moved demos**: explicit undefined checks with localized error throws. Project convention.
- **`ContentBlockDelta` signature field**: the optional `signature?: string` does not relax validation for sibling delta types. Verified.
- **Single-block thinking limit**: documented inline; multi-block + redacted handling is INTR-79's scope (description updated this round to reference commit `c1b0d83` as partial progress).

### Test results

- `make all` passes: 1554 tests (1545 main + 9 separated)
- Live-API smoke tests pass against the new model IDs (quickstart, multi-provider cheap+smart, coding-agent)
- Both moved demos run end-to-end (mechanically — the demo's user-INBOX-shutdown trigger is a pre-existing bug unrelated to this branch, worth a follow-up)

PR #39, opened 2026-05-26
Refactor harness connector routing into a testable unit

alexanderguy:
## Summary

- Extract the connector-thread routing decision out of `createHarness` into a self-contained `createConnectorRouter` unit with a pure `route()`, a separate `commit()`, and a matched `snapshot()`/`restore()` pair for storage round-trips.
- Drive `createHarness` through the router; delete the four loose state bindings, the partial-state guards, and the three inline helpers they were paired with.

## Changes

- `packages/harness/src/connector-router.ts`: new — router factory, `RouteDecision` discriminated union, `ConnectorReplyParts`, `NoActiveConnectorThreadError`.
- `packages/harness/src/connector-router.test.ts`: new — 15 unit tests covering inbound start/continue/passthrough, outbound `composeReply` with and without an active thread, `onReplySent` advance, snapshot/restore round-trip, and defensive copying.
- `packages/harness/src/harness.ts`: switch INBOX watch and `connector.reply` reactor-event handler to the router; remove inline routing helpers; route persistence through `router.snapshot()`.
- `packages/harness/src/index.ts`: re-export router factory, types, and the named error class.

## Testing

- `make all` clean against HEAD: 2238 + 9 tests pass, 0 fail.
- Each commit passes `make all` standalone (bisect-friendly).

## Behavior delta

One intentional, observable change, documented in commit 2: the `connector.reply` outbound handler previously no-opped silently when state was missing; it now throws `NoActiveConnectorThreadError`, which the existing fire-and-forget IIFE catches and surfaces via `logger.error`. There is no legitimate path where `connector.reply` fires without an active thread, so loud-log is the correct shape.

Closes INTR-76

### DIFF
--- a/packages/harness/src/connector-router.test.ts
+++ b/packages/harness/src/connector-router.test.ts
@@ -1,0 +1,301 @@
+import { describe, test, expect } from "bun:test";
+
+import { createInboundMessage } from "@intx/mime";
+import type { ConnectorThreadState, InboundMessage } from "@intx/types/runtime";
+
+import {
+  createConnectorRouter,
+  NoActiveConnectorThreadError,
+} from "./connector-router";
+
+function startMessage(opts?: { subject?: string }): InboundMessage {
+  return createInboundMessage({
+    from: "user@example.com",
+    to: ["agent@example.com"],
+    content: "hello",
+    messageId: "<root@example.com>",
+    ...(opts?.subject !== undefined ? { subject: opts.subject } : {}),
+  });
+}
+
+function continuationByReferences(
+  threadRoot: string,
+  opts?: { from?: string; messageId?: string },
+): InboundMessage {
+  return createInboundMessage({
+    from: opts?.from ?? "user@example.com",
+    to: ["agent@example.com"],
+    content: "more",
+    messageId: opts?.messageId ?? "<reply-1@example.com>",
+    references: [threadRoot],
+  });
+}
+
+function continuationByInReplyTo(
+  lastMessageId: string,
+  opts?: { from?: string; messageId?: string },
+): InboundMessage {
+  return createInboundMessage({
+    from: opts?.from ?? "user@example.com",
+    to: ["agent@example.com"],
+    content: "more",
+    messageId: opts?.messageId ?? "<reply-2@example.com>",
+    inReplyTo: lastMessageId,
+  });
+}
+
+function unrelatedMessage(): InboundMessage {
+  return createInboundMessage({
+    from: "stranger@example.com",
+    to: ["agent@example.com"],
+    content: "unrelated",
+    messageId: "<other@example.com>",
+  });
+}
+
+describe("createConnectorRouter", () => {
+  describe("route + commit (inbound)", () => {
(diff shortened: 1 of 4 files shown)


alexanderguy:

Reviewed `origin/main...HEAD` as it stands. Two commits, four files.

### Commit-message audit

Ran against `git log origin/main..HEAD`:

- Subjects over 72 chars (`awk 'length > 72'`): none.
- Body lines over 72 chars: none.
- Prefix scan (`grep -E '^(feat|fix|chore|docs|refactor|test|WIP|wip|\[)'`): none.
- Filename-in-subject scan (`grep -E '\.(ts|md|json|js|tsx|py)([ ":]|$)'`): none.
- Trailing punctuation: none.
- External tracker IDs in bodies (`grep -nE '(INTR-|JIRA-|#[0-9]+|Closes [A-Z]+-)'`): none. `Closes INTR-76` lives only in the PR body.
- Forward/backward commit references (`grep -niE '(previous commit|next commit|earlier commit|as discussed)'`): none.

### Commit coherence

- `34395c1` — touches `connector-router.ts`, `connector-router.test.ts`, `index.ts`. Body describes the router introduction and the barrel export.
- `361daac` — touches `harness.ts`. Body describes the call-site switch, the helper deletions, the synchronous-commit ordering, and the outbound silent-drop → loud-log delta.

Both diffs match their messages. Each commit passes `make all` standalone.

### What the branch does

`createHarness` no longer owns the connector-thread routing decision. A `createConnectorRouter` unit in `@intx/harness` holds the singleton thread state, classifies inbound messages (`route()` → `start` / `continue` / `passthrough`), advances state on commit, and produces the threading headers for outbound replies (`composeReply()`, throwing `NoActiveConnectorThreadError` when no thread is active). `snapshot()` / `restore()` form a matched pair for persistence round-trips through the existing context-store cadence.

The INBOX watch handler asks the router for a decision, commits router state synchronously before any await (matching the prior synchronous-mutation property), then delivers and consumes. The reactor-event `connector.reply` handler asks the router for the reply headers inside the existing fire-and-forget IIFE; missing-state firings now surface via `logger.error` rather than no-op silently.

### Behavior delta

One, documented in commit 2: a `connector.reply` event with no active thread previously dropped silently. It now throws into the IIFE's `try/catch` and surfaces via `logger.error`. There is no legitimate path where `connector.reply` fires without an active thread, so loud-log is the correct shape.

### Findings

None.

### Cited checks

- Commit-message audit: the `git log ... | awk` and `git log ... | grep` commands listed above, each run against `origin/main..HEAD`.
- `make all` clean on HEAD: `bun test` reports 2238 + 9 pass, 0 fail; `bun run lint` clean; `tsc -b --noEmit` clean.
- `make all` clean on each commit standalone: confirmed during the rebase that produced this history (each commit was a stop point with `make all` run before `--continue`).
- Behavior parity: read `git diff origin/main...HEAD -- packages/harness/src/harness.ts` against the prior inline routing. Verified subject persistence on continue, `replyTo` advance on continue, and the synchronous-commit-before-await ordering. Did not audit harness callers outside the diff.

PR #40, opened 2026-05-26
Thread user mail into the agent's active connector when present

alexanderguy:
## Summary

- The connector is one durable shared thread per agent. Anyone who sends conversational mail to the agent — a human via a hub session, a parent agent that launched this one as a sub-agent, a peer agent — joins the thread. Participants accumulate; no one is displaced.
- When the hub composes mail to an agent and a connector thread is active, the route handler stamps `inReplyTo` and `references` from the cached state so the harness routes the message as a continuation. No per-participant gate at the comparison site: anyone with a session joins whatever thread is in progress.
- `connector.reply` outbound replies address the whole thread (`to: replyTo, cc: rest`) so the agent's response reaches every participant.

## Changes

- `packages/harness/src/connector-router.ts`: multi-participant router. `ConnectorThreadState` carries `replyTo` (most recent speaker — primary recipient) and `cc: string[]` (every other participant who has spoken, deduplicated, in arrival order). `continue` moves the prior `replyTo` into `cc` and sets `replyTo` to the new sender. `composeReply()` returns `{ to: replyTo, cc, inReplyTo, subject? }`. `onStateChanged` fires on any real state change (including `cc`); the router catches and logs subscriber throws so the call chain that invoked `commit`/`onReplySent` is never aborted.
- `packages/mime/src/mime.ts`: new `extractAddrSpec` utility — strips display name and angle brackets, lowercases, throws on inputs it cannot parse safely (quoted local-parts, trailing content after `>`, multiple `@`).
- `packages/harness/src/harness.ts`: watch callback wraps `route()` in try/catch so malformed `From:` headers log and fall through to passthrough.
- `packages/types/src/runtime.ts`: `ConnectorThreadState` promoted to an arktype carrying `cc: string[]`.
- `packages/types/src/sidecar.ts`: new `connector.state.changed` wire frame; uses the arktype directly so the `cc` field flows through.
- `packages/hub-sessions/src/ws/sidecar-handler.ts`: per-agent `connectorStates` map. Cache writes gated on `addressIndex.get(addr) === ws` ownership. Cache eviction inline with `addressIndex.delete` and at all live-takeover sites (`handleRegister` same-ws, ghost cleanup, `handleReconnect` post-challenge verify). New `getConnectorState(addr)` accessor.
- `packages/hub-api/src/routes/instances.ts`: POST `/:instanceId/mail` stamps threading headers from the cache whenever a connector exists — no `replyTo` comparison, no address normalization in the route handler.
- `packages/storage-isogit/src/store.ts`: `ConnectorThreadStateSchema` requires `cc: "string[]"`. Existing on-disk `metadata.json` files written under the previous shape will fail validation and need `bin/db-reset --clean` (acceptable under the current dev-state contract).
- `apps/sidecar/{main,session-manager,ws-client}.ts`: wire the harness `onConnectorStateChanged` option through to the wire frame sender.
- `docs/HARNESS_DESIGN.md`: section "Connector threads and user sessions" describes the shared-thread model, the route handler's precedence rule (session history > connector cache > none), and the two cache-empty windows during reconnect.

## Testing

- `make all` clean against HEAD: 2281 + 9 tests pass, 0 fail.
- Each commit passes `make all` standalone.
- Router tests cover: `cc` accumulation on `continue` from a different sender, dedup on re-entry, multi-sender accumulation in arrival order, subject preservation across many continues, `composeReply` returns a copy of `cc` (not the live array), `cc` carried through `onReplySent`, snapshot/restore round-trip with `cc`, `restore` takes a defensive copy of the input `cc`, `onStateChanged` fires when only `cc` changes, subscriber throws don't propagate out of `commit()` or `onReplySent()`, `route()` throws on unparseable `From:` headers on both `start` and `continue` branches, `replyTo` normalization strips display name and lowercases.
- Route-handler tests cover: no active connector → no threading headers; active connector started by the same user → user continues; active connector started by another peer → user joins the same thread (same headers stamped); session history takes precedence over the connector cache.
- Wire-path tests cover: `connector.state.changed` populates the cache and is readable via `getConnectorState`; the frame emits the matching event; disconnect clears the cache; live takeover via register evicts the prior owner's cached state.
- Storage tests cover: `ConnectorThreadState` with non-empty `cc` and `subject` round-trips through `writeMetadata` + `commit` + `load`; state with empty `cc` and no `subject` also round-trips.
- `extractAddrSpec` tests cover the accepted inputs (quoted display name, bare display name, angle brackets only, bare addr-spec, mixed-case lowercased, trimmed whitespace) and rejected ones (empty, no `@`, missing local-part, missing domain, trailing content after `>`, quoted local-part, multiple `@`).

## Behavior delta

When a user sends mail to an agent that already has an active connector thread — regardless of who started it — the message now threads as a continuation and the user becomes a participant. Today it would route as passthrough and the conversation would silently break. The agent's next `connector.reply` reaches every participant on the thread (most recent speaker as `to:`, rest as `cc:`).

## Out of scope

- Auto-close, idle timeout, or explicit clear of the connector thread. Threads persist for the lifetime of the agent.
- Removing participants from the thread when a session ends. `cc` only grows.
- Migration of existing on-disk `ConnectorThreadState` snapshots; old snapshots may be wiped via `bin/db-reset --clean`.
- Per-peer connector threads. The shared-thread model intentionally maintains one thread per agent.

Closes INTR-77

### DIFF
--- a/apps/sidecar/src/main.ts
+++ b/apps/sidecar/src/main.ts
@@ -1,6 +1,6 @@
 import { setup, getLogger } from "@intx/log";
 import { createInMemoryTransport } from "@intx/mail-memory";
-import type { InferenceEvent } from "@intx/types/runtime";
+import type { ConnectorThreadState, InferenceEvent } from "@intx/types/runtime";
 
 import { createSessionManager } from "./session-manager";
 import { createWsClient } from "./ws-client";
@@ -41,12 +41,21 @@ let forwardEvent: (a: string, s: string, e: InferenceEvent) => void = (
 ) => {
   // Replaced after ws-client is constructed.
 };
+let forwardConnectorState: (
+  a: string,
+  state: ConnectorThreadState | null,
+) => void = (_a, _state) => {
+  // Replaced after ws-client is constructed.
+};
 
 const sessions = createSessionManager({
   transport,
   dataDir,
   onEvent(agentAddress, sessionId, event) {
     forwardEvent(agentAddress, sessionId, event);
+  },
+  onConnectorStateChanged(agentAddress, state) {
+    forwardConnectorState(agentAddress, state);
   },
 });
 
@@ -59,6 +68,7 @@ const client = createWsClient({
 });
 
 forwardEvent = client.sendEvent;
+forwardConnectorState = client.sendConnectorState;
 client.connect();
 
 log.info("Sidecar {sidecarId} connecting to {hubUrl}", { sidecarId, hubUrl });
--- a/apps/sidecar/src/session-manager.ts
+++ b/apps/sidecar/src/session-manager.ts
@@ -38,6 +38,7 @@ import {
 import type { InMemoryTransport } from "@intx/mail-memory";
 import type { GrantRule } from "@intx/types/authz";
 import type {
+  ConnectorThreadState,
   InboundMessage,
   InferenceEvent,
   InferenceSource,
@@ -74,10 +75,16 @@ export type SessionEventSink = (
   event: InferenceEvent,
 ) => void;
 
+export type ConnectorStateSink = (
+  agentAddress: string,
+  state: ConnectorThreadState | null,
+) => void;
+
 export type SessionManagerConfig = {
   transport: InMemoryTransport;
   dataDir: string;
   onEvent: SessionEventSink;
+  onConnectorStateChanged: ConnectorStateSink;
 };
 
 // Sanitize an agent address into a safe directory name.
@@ -187,7 +194,7 @@ export function buildToolDispatch(
 export function createSessionManager(
   config: SessionManagerConfig,
 ): SessionManager {
-  const { transport, dataDir, onEvent } = config;
+  const { transport, dataDir, onEvent, onConnectorStateChanged } = config;
   const sessions = new Map<string, AgentSession>();
   const provisioned = new Map<string, ProvisionedAgent>();
   const pending = new Set<string>();
@@ -375,6 +382,9 @@ export function createSessionManager(
             lastCheckpointHashes.set(agentAddress, event.data.checkpointHash);
           }
           onEvent(agentAddress, sessionId, event);
+        },
+        onConnectorStateChanged(state) {
+          onConnectorStateChanged(agentAddress, state);
         },
       });
 
--- a/apps/sidecar/src/ws-client.ts
+++ b/apps/sidecar/src/ws-client.ts
@@ -28,7 +28,11 @@ import {
 } from "@intx/types/sidecar";
 import type { KeyPair } from "@intx/types/runtime";
 
-import type { SessionManager, SessionEventSink } from "./session-manager";
+import type {
+  ConnectorStateSink,
+  SessionManager,
+  SessionEventSink,
+} from "./session-manager";
 import { createPackReceiver, chunkPack } from "@intx/pack-transport";
 import { hexDecode, hexEncode } from "@intx/types";
 
@@ -49,6 +53,7 @@ export type WsClient = {
   connect(): void;
   close(): void;
   sendEvent: SessionEventSink;
+  sendConnectorState: ConnectorStateSink;
 };
 
 export function createWsClient(config: WsClientConfig): WsClient {
@@ -661,7 +666,18 @@ export function createWsClient(config: WsClientConfig): WsClient {
     });
   };
 
-  return { connect, close, sendEvent };
+  const sendConnectorState: ConnectorStateSink = (
+    agentAddress,
+    connectorState,
+  ) => {
+    send({
+      type: "connector.state.changed",
+      agentAddress,
+      connectorState,
+    });
+  };
+
+  return { connect, close, sendEvent, sendConnectorState };
 }
 
 function uint8ArrayToBase64(bytes: Uint8Array): string {
--- a/bun.lock
+++ b/bun.lock
@@ -255,11 +255,9 @@
       "dependencies": {
         "@intx/inference": "workspace:*",
         "@intx/log": "workspace:*",
+        "@intx/mime": "workspace:*",
         "@intx/types": "workspace:*",
         "arktype": "catalog:",
-      },
-      "devDependencies": {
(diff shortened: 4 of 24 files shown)


alexanderguy:

### What the branch does

The connector is one durable thread per agent — a multi-participant conversation that anyone can join (a human via a hub session, a parent agent that launched this one as a sub-agent, a peer agent that initiates a conversation). Participants accumulate as they speak; no one is displaced.

When the hub composes mail to an agent and a connector thread is active, the route handler stamps `inReplyTo` and `references` from the cached state so the harness routes the message as a continuation. No per-participant gate at the comparison site: anyone with a session joins whatever thread is in progress, including a thread started by an agent reaching out to this one.

The mechanism:

- **Router (`packages/harness/src/connector-router.ts`).** `ConnectorThreadState` carries `replyTo` (most recent speaker — primary recipient of the next outbound reply) and `cc: string[]` (every other participant who has spoken, deduplicated, in arrival order). The `continue` branch moves the previous `replyTo` into `cc` and sets `replyTo` to the new sender. `composeReply()` returns `{ to: replyTo, cc, inReplyTo, subject? }`. The `onStateChanged` callback fires on every real state change (any field including `cc`); the router catches and logs subscriber throws so the call chain that invoked `commit`/`onReplySent` is never aborted by a misbehaving observer.
- **Addr-spec normalization (`packages/mime/src/mime.ts`).** A new `extractAddrSpec` utility strips display name and angle brackets and lowercases the result. Inputs the simple parser cannot handle safely (quoted local-parts, content after `>`, multiple `@`) throw explicitly. The router calls it when storing `replyTo` so outbound `to:`/`cc:` headers are bare addr-specs.
- **Harness watch callback (`packages/harness/src/harness.ts`).** Wraps `route()` in try/catch so a malformed `From:` header on the wire logs and falls through to passthrough.
- **Wire frame (`packages/types/src/sidecar.ts`).** `connector.state.changed` carries the full `ConnectorThreadState` (now an arktype in `@intx/types/runtime`, so the schema flows through automatically).
- **Hub cache (`packages/hub-sessions/src/ws/sidecar-handler.ts`).** Per-agent `connectorStates` map. Cache writes gated on `addressIndex.get(addr) === ws` ownership. Cache eviction inline with `addressIndex.delete` under the same guard, and at all live-takeover sites (`handleRegister` same-ws re-register, ghost cleanup of a different ws, `handleReconnect` post-challenge verify). Bootstrap rides on the router's `restore()` firing the callback when persisted state transitions from cold-start `null` to a populated value.
- **Route handler (`packages/hub-api/src/routes/instances.ts`).** POST `/:instanceId/mail` stamps threading headers from the cache whenever the connector exists. No `replyTo` comparison, no address normalization in the route handler — anyone with a session joins the existing thread.
- **Storage (`packages/storage-isogit/src/store.ts`).** `ConnectorThreadStateSchema` requires `cc: "string[]"`. Existing on-disk `metadata.json` files written under the previous shape will fail validation and need `bin/db-reset --clean`; acceptable under the current dev-state contract.
- **Docs (`docs/HARNESS_DESIGN.md`).** Section "Connector threads and user sessions" describes the shared-thread model, the route handler's precedence rule (session history > connector cache > none), and the two cache-empty windows during reconnect.

### Affirmative claims and what proves them

- **Build clean:** `make all` exits 0 on HEAD. Each commit passes `make all` standalone (verified during rebase).
- **Test counts:** 2281 unit tests pass / 2 pre-existing skips / 0 fail; 9 integration tests (`tests/sidecar/deploy-flow.test.ts`, `tests/inference/transform-cutover.test.ts`) pass / 0 fail.
- **Shared-thread router contract (`connector-router.test.ts`):** `start` initializes `cc = []`; `continue` from the same sender keeps `cc` empty and advances `lastMessageId`; `continue` from a different sender moves the prior `replyTo` into `cc`; multiple continues accumulate participants in arrival order; `continue` from a sender already in `cc` does not duplicate them; `subject` is preserved across many continues; `composeReply` returns `{ to, cc, inReplyTo, subject? }`; `composeReply` returns a copy of `cc` (mutation does not affect state); `onReplySent` preserves `cc` through the `lastMessageId` advance; snapshot/restore round-trips `cc`; `restore` takes a defensive copy of the input including its `cc`; `onStateChanged` fires when `cc` changes even if other fields do not; subscriber throws do not propagate out of `commit()` or `onReplySent()`; `route()` throws on unparseable `From:` headers on both `start` and `continue` branches; `replyTo` normalization strips display name and lowercases.
- **Threading-policy correctness (`instances.test.ts`):** (a) no active connector → no threading headers, (b) active connector started by the same user → user continues the thread, (c) active connector started by another peer → user joins the same thread (same headers stamped), (d) session history takes precedence over the connector cache.
- **Wire-path integration (`sidecar-handler.test.ts`):** `connector.state.changed` populates the cache and is readable via `getConnectorState`; the frame emits the matching event; disconnect clears the cache; live takeover via register evicts the prior owner's cached state.
- **Storage round-trip (`storage-isogit/src/store.test.ts`):** `ConnectorThreadState` with non-empty `cc` and `subject` round-trips through `writeMetadata` + `commit` + `load`; state with empty `cc` and no `subject` also round-trips.
- **`extractAddrSpec` strictness (`mime/src/index.test.ts`):** accepted inputs (quoted display name, bare display name, angle brackets only, bare addr-spec, mixed-case lowercased, trimmed whitespace) and rejected inputs (empty, no `@`, missing local-part, missing domain, trailing content after `>`, quoted local-part, multiple `@`).
- **Lint/format/docs:** `make lint` exits 0; `tsc -b --noEmit` exits 0; the API-docs freshness check passes.

### Out of scope

- Auto-close, idle timeout, or explicit clear of the connector thread. Threads persist for the lifetime of the agent.
- Removing participants from the thread (e.g. when a session ends). `cc` only grows.
- Migration of existing on-disk `ConnectorThreadState` snapshots — the storage schema now requires `cc`, and old snapshots may be wiped via `bin/db-reset --clean`.
- Per-peer connector threads. The shared-thread model intentionally maintains one thread per agent.

PR #43, opened 2026-05-26
Install a default stderr sink at @intx/log module load

alexanderguy:
## Summary

- `@intx/log` now installs a default sync console sink at module load when no LogTape config is present, so diagnostics emitted before `setup()` is called no longer silently disappear.
- `setup()` passes `reset: true` so its `configure()` cleanly replaces the default — and any prior `setup()` call — without throwing.
- Both `@intx/log` and `@intx/log/hono` entry points side-effect-import the installer so consumers of either get the default behavior.

## Changes

- `packages/log/src/default-sink.ts`: new module. `installDefaultConsoleSink()` returns early when a config already exists; otherwise registers `getConsoleSink({ formatter })` with `ansiColorFormatter` (non-production) or `getJsonLinesFormatter()` (production), threshold `warning`. Called at module load.
- `packages/log/src/index.ts`: side-effect import of `./default-sink` at the top; re-exports `installDefaultConsoleSink` for tests that need to reinstall after `resetSync()`.
- `packages/log/src/hono.ts`: side-effect import of `./default-sink` so the `/hono` entry point installs the default too.
- `packages/log/src/setup.ts`: pass `reset: true` to `configure()`; document the re-entry semantics in the JSDoc.
- `packages/log/src/default-sink.test.ts`: new tests — module-load capture, pre-setup routing, sink-swap on `setup()` with post-setup routing assertion, idempotency.
- `packages/log/src/default-sink-routing.test.ts`: new tests pinning the `warning` threshold and the `NODE_ENV`-driven formatter selection.

## Implementation notes

The issue's pseudocode suggested `getStreamSink({ stream: process.stderr, formatter })`, but LogTape 2.x's `getStreamSink` requires a Web `WritableStream` and returns an `AsyncDisposable` that `configureSync` rejects. `getConsoleSink` is sync, takes the same formatters, and routes through `console.warn`/`console.error` — which lands on stderr in Node and Bun and on the devtools console in browsers. Same default destination in every runtime, no `process.stderr` dependency.

## Testing

- [x] `make all` passes.
- [x] 10 new tests in `packages/log/src/` covering install, swap, idempotency, threshold, formatter selection.

Closes INTR-91

### DIFF
--- a/packages/log/src/default-sink-routing.test.ts
+++ b/packages/log/src/default-sink-routing.test.ts
@@ -1,0 +1,115 @@
+import { describe, test, expect, beforeEach, afterEach } from "bun:test";
+import { getLogger, resetSync, installDefaultConsoleSink } from "./index";
+
+// Locks down two design contracts of the module-load default sink that
+// default-sink.test.ts does not currently exercise:
+//
+//   1. Threshold is "warning" — info/debug must be suppressed; warning,
+//      error, and fatal must be routed.
+//   2. Formatter selection follows NODE_ENV — production yields JSON Lines,
+//      anything else yields ANSI-colored text.
+//
+// Both behaviors are contracts a future refactor could accidentally break
+// (e.g., bumping threshold to "info" or flipping the env heuristic).
+
+/* eslint-disable no-console -- intentional spies on console methods */
+describe("default console sink routing contract", () => {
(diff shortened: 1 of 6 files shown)


alexanderguy:
Self-review (`code-review` skill) returned clean — no findings on the current branch.

PR #44, opened 2026-05-27
Pluggable mechanical retry policy for runInference

alexanderguy:
## Summary

- `runInference` in `@intx/inference` wraps each call in a mechanical retry layer: per-attempt event buffer, a `RetryPolicy` consulted on `inference.error`, and one `inference.retry` event emitted between attempts carrying the failed attempt's number, the policy-chosen `delayMs`, and the classified error.
- `createDefaultRetryPolicy()` ships conservative defaults: up to 3 attempts with 500/1000 ms backoff on `retryable` and `timeout`; up to 3 attempts on `quota_exhausted` honoring `retryAfterMs ?? 1000 ms`; immediate abort on `credential_failure`, `context_overflow`, `fatal`, `aborted`, and `protocol_mismatch`.
- The caller-supplied `AbortSignal` short-circuits the retry delay so a 60-second `retryAfterMs` on a quota error wakes the wrapper immediately on cancellation rather than blocking the full minute.
- Caller-visible event seqs stay contiguous across retries: each attempt runs against a private allocator and the buffer is re-stamped from `opts.nextSeq` at flush, so a discarded attempt does not leave gaps in the consumer's seq stream.
- `Dependencies.scheduler` is now required and exposes a `now()` method alongside the existing `setTimeout`, giving the wrapper a monotonic time source for `RetrySituation.elapsedMs`. The production default uses `performance.now()`; the test harness wires `now()` to its virtual clock in both inert and timer-enabled modes.
- `InferenceOptions.retryPolicy?: RetryPolicy` lets callers override per call. The contract types (`RetryPolicy`, `RetrySituation`, `RetryDecision`) live in `@intx/types/runtime` and are re-exported from `@intx/inference` for ergonomic imports.
- The retry contract is documented in the inference docs alongside the existing event protocol and per-call timeout sections.

## Test plan

- [x] `make all` clean — 2331 pass / 2 skip / 0 fail across 138 files, plus 9/0 secondary integration suite.
- [x] Retry-policy behavioural suite at `tests/inference/retry.test.ts` covers default 3-attempt cadence, abort-only policy, unlimited-retry success on attempt N, `quota_exhausted` with and without `retryAfterMs`, every non-retryable category, custom-policy `attempt` reading, `inference.retry` payload, policy throws / rejected promises, scheduler-driven delay, signal short-circuit during the delay, abort-vs-retry seq invariants, and contiguous caller-visible seqs across a retry.
- [x] Scheduler unit + harness tests at `packages/inference/src/scheduler.test.ts` and `tests/inference/scheduler.test.ts` cover monotonicity, the inert-mode `setTimeout` no-op / `now()`-tracks-clock asymmetry, and timer firing under virtual clock.
- [x] Timeout suite at `tests/inference/timeout.test.ts` exercises the default policy's three-attempt retry on `timeout` via three single-use stalls plus abort-propagation per attempt.
- [x] Direct unit coverage of `createDefaultRetryPolicy` decisions per category.

Closes INTR-88.

### DIFF
--- a/docs/INFERENCE.md
+++ b/docs/INFERENCE.md
@@ -70,6 +70,10 @@ The harness is the shared infrastructure that all provider adapters build on:
 
 Provider adapters never touch SSE parsing, connection lifecycle, abort handling, or event emission. They translate request/response shapes. The harness does everything else.
 
+`runInference` wraps each call in a retry layer that buffers an attempt's events until the attempt terminates with `inference.done` or `inference.error`. On `inference.error` the wrapper consults a `RetryPolicy` (see §Retry Behavior); on a retry decision the wrapper discards the failed attempt's buffered events, emits one `inference.retry` event between attempts, and re-issues the underlying HTTP request after the policy-chosen delay. On `inference.done` or an abort decision the buffered events flush to the caller in order. The buffer is per-call and bounded by the size of one attempt's event stream.
+
+The buffer-and-flush model guarantees the caller sees a single clean event stream — exactly one `inference.start`, no orphaned partial deltas, no leaked `inference.error`s from attempts the policy chose to retry. The cost is that no events reach the caller until the wrapper knows the attempt's terminal shape: a consumer that uses `for await ... break` to abandon a call mid-stream will not observe partial deltas because the wrapper holds them in the buffer until commit. Consumers that need token-by-token partials must pin a custom non-buffering wrapper at the inference layer — no streaming-partials emission API exists today.
+
 ## Event Protocol
 
 All inference activity — streaming tokens, tool execution, reactor state changes — emits events on a single protocol. This protocol is what session channel subscribers receive. There is no internal event format.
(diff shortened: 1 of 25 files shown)


alexanderguy:

**Style audit (cited):**
- Subject lengths: `git log origin/main..HEAD --format='%s' | awk 'length > 72'` → empty.
- Body lengths: `git log origin/main..HEAD --format='%b' | awk 'length > 72'` → empty.
- Subject prefixes / filename references / trailing punctuation / vague subjects: none.
- Body external-tracker references / cross-commit references / PR-discussion references: none.
- `make all`: 2331 pass / 2 skip / 0 fail across 138 files; secondary 9/0.

**Behavioural contracts the test suite pins:**
- Buffer-and-flush gives the caller a single clean event stream per call — exactly one `inference.start`, no orphaned partial deltas, no leaked `inference.error`s from attempts the policy chose to retry.
- Caller-visible event seqs stay strictly monotonic across retries; a discarded attempt's seq allocations come from a private per-attempt counter that never reaches the caller.
- `AbortSignal` is honored at attempt entry (next attempt yields `inference.error` of category `aborted`) and mid-delay (the await races the scheduler against the abort listener; the still-pending counterpart is cancelled and the listener removed).
- Policy throws and rejected Promises coerce to abort; the original `inference.error` surfaces and the policy's exception is logged at `warn`.
- `Dependencies.scheduler` is required; `runInference` validates `deps.fetch` and `deps.scheduler.now()` at entry with a plain-Error message naming the missing field.
- Default policy: 3-attempt cap, 500/1000 ms backoff on retryable/timeout, `retryAfterMs ?? 1000 ms` flat on quota, immediate abort on the deterministic-failure categories. The switch is exhaustive against `InferenceError.category` at compile time.

PR #45, opened 2026-05-29
Extract @intx/hub-agent from apps/sidecar

alexanderguy:
## Summary

- Carves a new `@intx/hub-agent` package out of `apps/sidecar`, mirroring the shape of `@intx/hub-sessions`: per-agent disk layout (`AgentRepoStore`), Ed25519 key custody (`AgentKeyStore`), the harness lifecycle (`SessionManager`) behind a `HarnessBuilder` seam, the hub WebSocket protocol (`HubLink`), and a `SidecarOrchestrator` that wires the four together.
- `apps/sidecar` becomes a ~40-line entry point that supplies env vars, the default `HarnessBuilder` (composed of `@intx/crypto-node`, `@intx/tools-posix`, `@intx/tools-lsp`, `@intx/storage-isogit`, `@intx/authz`, `@intx/inference`, `@intx/harness`), and the low-level crypto primitives. Any future host can substitute its own.
- `CryptoProvider` grows a `signSSH` method. `@intx/types` gains `base64Encode`/`base64Decode`/`hasCode`. `@intx/harness` composes layered tool dispatch (mail + caller + deploy) inside `createHarness`. `@intx/mail-memory` exposes a `HubTransport` interface so the package never pins the concrete `InMemoryTransport` class.

Closes INTR-134

## Changes

- `packages/hub-agent/` — new package containing the extracted runtime.
- `packages/types/` — `base64.ts`, `has-code.ts`.
- `packages/harness/src/tools.ts` — layered dispatch with mail/caller/deploy collision checks.
- `packages/crypto-node/src/provider.ts` — `signSSH` method on `NodeCrypto`.
- `packages/mail-memory/src/transport.ts` — `HubTransport` interface.
- `packages/hub-sessions/src/agent-repo.ts` — drops its inline `hasCode` copy.
- `apps/sidecar/src/index.ts` — thin wiring; `default-harness.ts` is the only other source file.
- `tests/hub-agent/deploy-flow.test.ts` — integration test relocated and retargeted at the new entry point.

## Testing

- [x] `make all` green (build, lint, tests)
- [x] 2348 unit tests pass; integration test (4 tests) passes against the relocated sidecar entry point
- [x] Self-review via the `code-review` skill — punch list posted as a PR comment


alexanderguy:

Punch list from the whole-branch review against `origin/main`. Every finding describes the current state of the diff.

### Substantive findings

- **`HubLink` reconnect timer callback** — `packages/hub-agent/src/ws/hub-link.ts:677-686`. The scheduler callback re-checks `closed` before re-entering `connect()`. The close-side `cancelReconnect` is the load-bearing protection against real `setTimeout`; the in-callback re-check is one line of defense in depth at the injection boundary for fake schedulers whose cancel is a no-op.
- **`AgentKeyStore.fileExists`** — `packages/hub-agent/src/agent-key-store.ts:249-260`. The catch narrows to `ENOENT` only; any other failure mode (`EACCES`, `EBUSY`, `EIO`, …) re-throws, so a permission-denied existence check can't silently drop into the mint-and-persist branch and overwrite the agent's private key.
- **`HubLink.handleChallengeFailed`** — `packages/hub-agent/src/ws/hub-link.ts:319-325`. `destroySession` runs first and the catch logs any error; `keyStore.forgetAgent` runs only after the session-side teardown returns. Disposers and mail-commit code that close over the agent's crypto handle still see it during cleanup.
- **`scanKeys` / `scanConfigs` path assembly** — `packages/hub-agent/src/agent-key-store.ts:163-175`, `packages/hub-agent/src/agent-repo-store.ts:207`. Paths are built via `path.join` against `KEYS_DIR_NAME`, `PRIVATE_KEY_FILE`, `PUBLIC_KEY_FILE`, `META_FILE` exported from `packages/hub-agent/src/agent-paths.ts`. The on-disk filenames live in one place across the package.
- **`HubTransport` interface** — `packages/mail-memory/src/transport.ts:52-63`, re-exported from `packages/mail-memory/src/index.ts:1`. `SessionManager`, `HubLink`, `SidecarOrchestrator` accept `HubTransport` rather than the concrete `InMemoryTransport`; a custom host can supply its own backend without changing the package.
- **`default-harness` rollback** — `apps/sidecar/src/default-harness.ts:103-117`. Dispose failure is logged through `@intx/log`; the original construction failure still propagates on line 118.
- **`SessionManager.runDisposers`** — `packages/hub-agent/src/session-manager.ts:346-368`. Returns `Error[]`. `destroySession` and `abortSession` log a summary line when the array is non-empty; the aggregate is not thrown because session teardown runs from recovery paths where the original failure must remain visible.
- **`HarnessConfig.deployTools` docstring** — `packages/harness/src/config.ts:104-107`. Handlerless deploy tools that match no other source route to the declared-but-no-handler error. (The collision check at `packages/harness/src/tools.ts:639-642` blocks names that appear in both `tools.definitions` and `deployTools` at construction time, so the caller-runner path is only reachable for names declared in `tools.definitions`.)
- **Comments** — `packages/hub-agent/src/sidecar-orchestrator.ts:8-14`, `packages/harness/src/harness.ts:93-97`. Self-contained descriptions of the current responsibilities; no historical framing.

### Observations (informational, not fixed in this branch)

- **`agent-key-store.ts:184-189`** — `JSON.parse` failure, non-object parses, and missing/non-string `address` cases `continue` silently. `agent-repo-store.ts:218` does log via the arktype `validated instanceof type.errors` path; the key-store side could match that for symmetry. Worth a follow-up issue.
- **`HubTransport.deliver` semantic** — the JSDoc says "drop into the address's inbox" but does not formalize that the recipient must be registered first. `InMemoryTransport.deliver` throws if it isn't; the interface comment could note that.

### Out of scope for this PR

- `await has no effect` lint warnings in `packages/hub-agent/src/ws/hub-link.test.ts` — pre-existing.
- `@intx/hub-sessions`' `AgentRepoStore` naming collision with `@intx/hub-agent`'s `AgentRepoStore` — informative parallel; test sites that import both alias one of them.
- `isomorphic-git` catalog migration — eligible for a small cleanup commit later.
- `@intx/crypto-node` as a `devDependency` of `@intx/hub-agent` — the runtime barrel never re-exports from crypto-node and no non-test source imports it; the runtime seam holds.

### Test results

- `make all`: build clean, lint clean, 2348 tests pass / 2 skip / 0 fail (140 files); the relocated integration test passes (4 / 4).

PR #47, opened 2026-06-02
Materialize assets on agents via synthetic hub-hosted git repos

alexanderguy:
## Summary

- Introduces a kind-keyed asset substrate alongside agent state, the first concrete kind being `skill` (markdown + sibling files per the agentskills.io spec, validated at hub-side write).
- Adds per-agent attachment, session-start materialization on the sidecar, and an `<available_skills>` stanza in the agent system prompt.
- Generalizes the sidecar pack frame protocol from a single agent-state-specific shape to `repo.pack.*` with `(direction, repoId, ref, pack)` addressing and an optional `mountPath` for asset materialization.

### Pack frame protocol

The four sidecar pack frames are renamed to `repo.pack.{push,done,ack,reject}` and each carries a `repoId: { kind, id }` alongside the existing `agentAddress`. For the only kind in use today (`agent-state`), the two identifiers coincide; the duplication exists so a future asset attachment can decouple source-repo identity from destination-agent identity. `RepoPackDoneFrame` gains an optional `mountPath` — when set, the sidecar routes the pack to a workspace materialization path instead of the agent's deploy tree.

### Asset substrate

The kind-keyed RepoStore picks up a `skill` handler at `assets/skill/`. SKILL.md frontmatter is arktype-validated on every push (`name` lowercase-kebab regex up to 64 chars, not `anthropic` or `claude`; `description` 1–1024 chars with no XML tag pattern; `name` must equal the containing directory). The Claude Code optional-fields superset is accepted and ride-stored on a per-asset skill index but not acted on in v1.

The `KindHandler.validatePush` interface widens to receive a `readBlob` callback so kind handlers can inspect blob contents during validation. `RepoStore.writeTree` now runs `validatePush` symmetrically with `receivePack` so both write paths surface content-rejection errors with the same `path_violation:` prefix and the handler's reason string.

### Tables

- `asset (id, tenantId, kind, name, displayName, creatorPrincipalId, createdAt, updatedAt)` with `unique(tenantId, kind, name)` — mirrors the credential model's per-tenant scoping.
- `agent_asset (id, agentId, assetId, ref, accessMode, createdAt)` with `unique(agentId, assetId)` — prevents the same asset being attached to one agent twice.
- `session_asset (instanceId, agentAssetId, mountPath, assetPackSha, sourceCommitSha, materializedAt)` with PK on `(instanceId, agentAssetId)` and an index on `assetPackSha`. Schema comment documents the canonical reading: rows record sidecar acks during launch; multi-attachment partial-success leaves rows for the attachments that did ack.
- The legacy `agent.skills` JSONB column is dropped.

### Service layer

`createAssetService` exposes `createAsset` (rejects `agent-state`, validates asset names lowercase-kebab, runs `initRepo` before the row insert), `populateAsset` (routes through `writeTree` so kind-handler validation runs), `attachAsset` (translates SQLSTATE 23505 to `duplicate_attachment` and 23503 to `invalid_reference`), and `listAgentAssets` (ordered by `createdAt, id` for stable downstream consumers).

### Sidecar materialization

`applyAssetPack` is a free function on `@intx/hub-agent` that materializes a pack at `<workspaceRoot>/<mountPath>/` using a transient scratch git directory plus manual blob extraction — the workspace is plain files, not a git working tree. mountPath segments are rejected if entirely dots or if `path.posix.normalize` collapses them to escape the root. Submodule references throw `asset_materialization_failed: submodule reference at ... not supported`. Asset packs are unsigned in v1; validation lives on the hub-side write path.

`SessionManager.applyAssetPack(agentAddress, mountPath, pack, ref, commitSha)` mirrors `applyDeployPack` and delegates to the free function, resolving `workspaceRoot = path.join(repoStore.getAgentDir(addr), "workspace")`.

### Session start

`launchSession` gains an `instanceId` parameter. The flow:

1. Resolve attached assets up front.
2. For each `kind=skill` attachment, pull the per-asset skill index and accumulate `<available_skills>` entries.
3. Append the stanza to `deployContent.systemPrompt` before `writeDeployTree` so it lands in `deploy/prompt.md`.
4. Write deploy tree, send deploy pack.
5. For each attachment, resolve mountPath (`skills/<asset.name>/` for kind=skill; throws `mount_path_required:` for unhandled kinds), resolve the ref, build the pack, compute its sha256, insert the `session_asset` row before sending the pack, then send. The rollback path deletes the row on `sendPack` failure (with a logged warning on rollback failure, so the primary sendPack error reaches the caller).
6. Send session start.

### System prompt

`buildAvailableSkillsStanza` renders `<available_skills>` with one `<skill>` per entry: qualified name `<asset.name>/<skill-name>`, description, workspace path. Returns the empty string when no skills are attached. XML-special characters in name and description are escaped.

## Test plan

- [x] `make all` (lint + tsc -b + bun test) on the final commit: 2402 pass / 2 skip / 0 fail in the main suite, 10 pass / 0 fail in the serial pass.
- [x] `bin/db-reset` applies the migration chain from 0000 through 0026 against a fresh database cleanly.
- [x] In-process E2E at `tests/skill-attachment-flow.test.ts` exercises: synthetic skill asset creation + population + attachment, manual materialization via `applyAssetPack` against an `os.tmpdir()` workspace, manifest insert, and a harness-driven `runInference` with `read_file` tool dispatches asserting on (b) materialized SKILL.md + sibling content, (c) the `session_asset` row's `instanceId` / `mountPath` / `sourceCommitSha` / `assetPackSha`, and (a) the outbound system prompt containing the `<available_skills>` stanza.
- [x] Agent-state deploy + state pack flows preserved (no behavior change): existing deploy-flow integration test, agent-repo tests, hub-link tests, and store tests pass.

Closes INTR-95


alexanderguy:

Reviewer-of-record audits run in-session with own-eyes inspection of raw output.

### Reviewer-of-Record checks

- `git log --oneline origin/main..HEAD`: 12 commits, every subject a plain-English sentence with no prefix, no filename token, no trailing punctuation, no ticket ID.
- `git diff origin/main...HEAD --stat`: 70 files, +11784 / −840.
- Bin marker scan in `--stat`: `(no Bin markers)`. No source file is being treated as binary.
- Subject length filter `awk 'length > 72'`: empty.
- Body length filter `awk 'length > 72'`: empty.
- Subject prefix scan `grep -E '^([a-z]+:|\[|\()'`: `(none)`.
- Subject trailing punctuation scan `grep -E '[.!?]$'`: `(none)`.
- Body tracker scan `grep -nE '(INTR-[0-9]+|JIRA-[0-9]+|#[0-9]+|Closes |Fixes #)'`: `(none)` in commit bodies. The `Closes INTR-95` line lives in the PR description, not in commit messages.

### Build / test verification

- `make all` against the tip: 2402 pass / 2 skip / 0 fail in the main suite (146 files), 10 pass / 0 fail in the serial pass.
- `bin/db-reset` against a fresh database applies all 26 migrations cleanly.
- E2E `tests/skill-attachment-flow.test.ts` runs in-process against `os.tmpdir()` with a mocked-inference harness.

### Code-vs-message coherence (spot-checked)

For the two commits whose messages most directly describe constraint and routing behavior — `0cf2562` (asset service layer) and `43ef84a` (session_asset wiring) — I cross-referenced the message claims against the production code:

- `attachAsset`'s body claim "translates (agentId, assetId) duplicates into duplicate_attachment" matches `packages/hub-sessions/src/asset-service.ts` `attachAsset`'s SQLSTATE-23505 catch block, which corresponds to the `unique(agent_id, asset_id)` constraint defined in `packages/db/src/schema/agent-assets.ts`.
- `launchSession`'s body claim "resolves mountPath as `skills/<asset.name>/` for kind=skill (throwing `mount_path_required:` for unhandled kinds)" matches `packages/hub-sessions/src/session-service.ts` `resolveMountPath`'s switch — `case "skill"` returns the templated path; the `never` default and the `agent-state` arm both throw with the documented prefix.

### Surface area I did NOT exercise

- Multi-tenant assets at scale: tests use a single tenant + agent fixture. Cross-tenant collision behavior follows from the `unique(tenantId, kind, name)` constraint on `asset`, but I did not assert on it.
- Real-LLM inference path: the E2E mocks inference via `@intx/inference-testing`. No real provider was called.

### Verdict

Accepted.

PR #52, opened 2026-06-04
Split @intx/agent into AgentDefinition and AgentEnv; narrow @intx/harness

alexanderguy:
## Summary

- `@intx/agent` exposes `defineAgent` + `createAgent(def, env)`. The
  definition is portable hashable data (id, system prompt, tool
  factories, optional director ref, inference preferences,
  capabilities, tags); the env supplies the active inference source,
  context store, workdir, audit sink, authorize callback, and director
  registry. `audit`, `authorize`, and `directors` are required env
  fields with no read-site defaults; `@intx/agent/testing` ships no-op
  stand-ins for tests and examples. `defineTool` produces bundle-shape
  factories with package-namespaced ids and declared env-key
  requirements; `defineDirector` plus `DirectorRegistry` give
  workflow-author directors the same shape as built-ins.
  `canonicalizeForHash` produces stable deploy-hash bytes from a value
  tree, rejecting non-JSON values and cycles. `validateEnv` is
  presence-only and surfaces `AgentEnvError` listing missing keys and
  the tools or director that declared them. `getRequiredEnvKeys`
  returns a `RequiredEnvKeys` struct that names both the env-key
  surface and any unresolvable director id, so env-builder UIs answer
  both questions in one call.

- `@intx/harness` narrows to a composition layer over `@intx/agent`.
  `createHarness(def, env)` wraps `env.storage` with connector-state
  load/persist via a `Proxy` that overrides only `load` and
  `writeMetadata` and forwards everything else, calls
  `createAgent(def, env)` once, drains the agent's event stream to
  send `connector.reply` events via `env.transport`, and runs the
  INBOX watch loop. The reactor is wrapped exactly once, inside
  `createAgent`. The narrowed `Harness` surface is `close`,
  `deliver`, `setSource`, `stream`, and `blobReader`. `defineMailTools`
  wraps a mail-tool bundle as an `AnnotatedToolFactory<MailEnv>`
  declaring `requires: ["transport", "address"]`.

- The wrappedStorage's connector-state restore is boot-only: it
  subscribes to the router's `onStateChanged` and treats in-memory
  state as authoritative the moment `commit()` runs, so mid-cycle
  `contextStore.load()` calls no longer clobber the committed state.

- The `commitErrors` accumulator and flush hook live in `@intx/agent`.
  The agent intercepts `inference.error` and `reactor.error` events,
  accumulates `ErrorRecord`s, and flushes at the assembly's
  `afterCheckpoint` and `onShutdown` hooks. Concurrent flush
  invocations are serialized through a shared follow-up promise so
  overlapping hooks do not double-commit the same prefix.
  `readDeployTree` lives in `@intx/hub-agent` alongside the rest of
  the sidecar's agent-data substrate.

- The lock keys on `env.workdir`. `AgentContextLockError` exposes the
  resolved path via `.workdir`. `createAgent` and `createHarness` each
  release their resources (workdir lock; tool-bundle disposers;
  underlying agent) on every construction-failure path so a throw
  past `validateEnv` does not leak.

- Two fixtures live in `packages/agent/src/internal-fixtures/`. The
  planner-shape fixture instantiates against a bare `BaseEnv`. The
  mail-shape fixture exercises (a) `AgentEnvError` blame for a
  transport-less env and (b) success against a transport-bearing env.
  Each fixture's test counts director-factory invocations as a proxy
  for reactor-once: every `createAgent` invokes the resolved factory
  exactly once before passing it to the assembly. The composition
  layer carries its own reactor-once cross-check in
  `packages/harness/src/harness.test.ts`.

- `apps/sidecar/src/default-harness.ts` constructs the definition and
  env, calls the new `createHarness`, and threads the agent's event
  stream onto its existing `onEvent` callback. `@intx/hub-agent`'s
  session manager constructs the harness synchronously and tears it
  down via `close()`. Every `examples/agent-*`, `examples/coding-agent`,
  `examples/posix-demo`, `examples/ring-demo`, and `tests/agent/`
  consumer migrates to the new surface. READMEs for `@intx/agent`,
  `@intx/harness`, `@intx/tools-mail`, and the affected example
  packages match the current surface.

## Verification

- `make all` exits 0: 2790 pass / 0 fail / 2 skip in the main pass
  plus 40 pass / 0 fail in the secondary pass.
- The 2 skipped tests are pre-existing on `origin/main` (DB-env-gated
  integration tests) and are unrelated to this branch.
- Per-commit `tsc -b --noEmit` passes at every commit on the branch;
  the history bisects against build.
- `make lint` passes (prettier, eslint, API-docs check).

Closes INTR-154
Closes INTR-167
Closes INTR-169


alexanderguy:

The branch went through the `code-review` skill's loop until two
consecutive rounds returned clean. Test coverage and known-design
exceptions documented below.

### Test coverage

The full workspace suite (`make all`) is 2790 pass / 0 fail / 2 skip.
The two skips are pre-existing on `main` (DB-env-gated integration
tests). The new code lands its own load-bearing tests:

- `packages/agent/src/agent.test.ts` pins `createAgent`'s lock
  release across every construction-failure path, the intra-
  `resolveTools` bundle-dispose rollback under `DuplicateToolError`,
  the post-`resolveTools` bundle-dispose rollback under a
  director-factory throw, and the async-disposer-rejection swallow on
  both rollback paths.
- `packages/agent/src/env-validation.test.ts` covers per-contributor
  blame for missing keys, `BaseEnv`-only blame when `directors`
  itself is absent, the `unresolvedDirectors` channel on
  `AgentEnvError`, `getRequiredEnvKeys` partial-result + unresolved-
  id surfacing in a single call, and `effectiveDirectorRef`'s
  absent-director normalization.
- `packages/agent/src/flush-errors.test.ts` and
  `packages/agent/src/audit-integration.test.ts` cover the error
  accumulator's splice-only-on-success semantics, the shared
  follow-up serialization across concurrent flush hooks, and the
  per-event-shape recording.
- `packages/agent/src/internal-fixtures/{planner,mail}.test.ts`
  pin reactor-once by counting director-factory invocations through
  a fixture proxy.
- `packages/harness/src/harness.test.ts` carries the composition-
  layer reactor-once cross-check, the boot-only-restore dirty-bit
  semantics through the directly-importable
  `createWrappedStorageOverrides`, the await-the-callback contract
  for `invokeReplySendFailed` / `invokeReplyDrainTerminated`, INBOX
  watch subscription and unsubscribe, and the workdir-lock collision
  case.
- `packages/agent/src/{canonicalize,namespace,tool,director,director-registry,default-director,default-director-decide,source,send-queue,stream,lock}.test.ts` carry the per-primitive suites.
- `tests/agent/{lifecycle,send-flow}.test.ts` cover the end-to-end
  reactor cycle against `setupHarness`-driven inference stubs.

### Known-design exceptions

These are intentional design choices documented in code and
authorized via Greybeard rulings during review. Each is documented
at its call site so a future reader sees the rationale alongside
the code, not buried in PR history.

- `defineMailTools` declares `requires: ["transport", "address"]`
  even though `address` is consumed by `createHarness` (logging-
  only), not the factory body. Authorized so `validateEnv` blames
  the mail factory at construction rather than letting the watch
  loop discover the gap under operational load.
  (`packages/harness/src/harness.ts:184-243`)
- `invokeReplySendFailed`, `invokeReplyDrainTerminated`, and
  `createWrappedStorageOverrides` are exported from
  `packages/harness/src/harness.ts` "for the regression test in
  this package" via docstring convention rather than a separate
  testing entry-point. Authorized because the helpers are tightly
  coupled to the production contract they exercise.
- `let sendQueue: SendQueue<...>` in `packages/agent/src/agent.ts`
  carries an `eslint-disable prefer-const` for an irreducible
  forward-declaration cycle (`handleEvent` reads `sendQueue` from
  closure; `sendQueue.start` calls `reactor.deliver`; `reactor` is
  constructed with `onEvent: handleEvent`). Authorized; the
  alternative breaks at one of the three edges with no payoff.

### Follow-ups filed

- INTR-170 (`@intx/tools-posix` env-DI working directory) — gated on
  a design decision about whether the posix tool reads from
  `env.workdir` or a new env field. Out of scope for this PR per
  Greybeard's earlier ruling.

PR #53, opened 2026-06-05
Wire compactors through createAgent and expose them to directors

alexanderguy:
## Summary

- `BaseEnv` carries an optional `compactors?: Record<string, Compactor>` field that `createAgent` threads into the reactor assembly.
- The director factory's `agentContext` carries a `compactorNames: readonly string[]`, listing the names the deployer registered, so a director picks a known name to pass to `caps.compact(name, reason)`.
- A new integration test drives `caps.compact("name", reason)` against a registered recording compactor and asserts the strategy's `apply()` runs with the inbound user turn and the director's trigger; a regression test pins the reactor's existing fatal-error behavior for unregistered names.

## Verification

- `make all` exits 0 against the full project graph: lint (prettier + eslint + docs-gen check) passes, `tsc -b --noEmit` passes, the main test pass runs 2768 tests with 0 failures, and the secondary-pass tests run 14 more with 0 failures.
- The new tests live in `packages/agent/src/compactor-wiring.test.ts` (end-to-end + regression) and `packages/agent/src/env-validation.test.ts` (the `compactors`-is-optional contract); they run as part of `make all`.

Closes INTR-171

### DIFF
--- a/packages/agent/README.md
+++ b/packages/agent/README.md
@@ -57,6 +57,56 @@ const { reply } = await agent.send("hello");
 await agent.close();
 ```
 
+## Compactors
+
+A compactor rewrites the conversation history when the director asks for
+it. The deployer registers compactors on `env.compactors` keyed by name;
+the director picks a name at construction and emits
+`caps.compact(name, reason)` when it wants the history rewritten.
+
+```ts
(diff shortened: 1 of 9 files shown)


alexanderguy:

PR #55, opened 2026-06-09
Tenancy-inheritable assets and the package-registry asset kind

alexanderguy:
Implements [INTR-178](https://linear.app/abklabs/issue/INTR-178/tenancy-inheritable-assets-and-the-package-registry-asset-kind).

## Summary

- Introduces `packages/db/src/asset-resolution.ts` with `resolveAssetByName`, `resolveAssetById`, and `listAssetsForTenant` — the tenant ancestor walker for the `asset` table. Mirrors the existing credential and provider walkers (child shadows parent, MAX_DEPTH=20 cycle guard, dedup by `(kind, name)` with inner overriding outer and origin tags on every row).
- Adds `GET /tenants/:id/assets?kind=...&inherited=...` and `GET /assets/:id` to the hub API. The list endpoint walks the ancestor chain when `inherited=true` (default) and tags each row with the tenant that supplied it; the lookup endpoint validates chain membership before returning. Mirrors the `providers?inherited=true` shape.
- Schema unchanged. The `asset` table already encoded the pattern; this PR adds the walker, the routes, and tests.

Two commits:

1. `Add the tenant-aware asset resolver` — the walker plus 511 lines of tests covering walk order, MAX_DEPTH, override semantics across both `skill` and `package-registry` kinds, and origin tagging.
2. `Expose tenant-aware asset listing and lookup over the hub API` — the routes, their tests, the OpenAPI regeneration, and the `AssetResponse` / `AssetWithOriginResponse` arktype shapes.

## Test plan

- [x] `make all` green on the rewritten branch (2820 tests pass, 0 fail)
- [x] Walker tests cover walk order, MAX_DEPTH cycle guard, `(kind, name)` collision override, no-orphan
- [x] Hub-API tests cover attach-at-root-tenant + list-from-child-agent + observe inheritance and override
- [x] Existing credential and provider tests still pass
- [x] `bin/gen-api-docs.ts --check` passes (OpenAPI in sync)

## Out of scope

- Sidecar materialization. Triggered by deploy-pack contents, which are written by INTR-108's resolver. INTR-108 rebases on top of this PR.
- A `kind`-narrowing type. The schema column stays `text`; consumers pass `kind: string` for now. A future ticket can narrow the column type, the walker signatures, and the route handler params in one coherent change once a consumer load-bears against it.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -103,7 +103,9 @@
 | GET | /api/tenants/:tenantId/agents/:agentId/history/:ref | Show changes in a commit |
 | GET | /api/tenants/:tenantId/agents/:agentId/branches | List branches |
 | POST | /api/tenants/:tenantId/agents/:agentId/history/:ref/restore | Restore agent data to a previous state |
+| GET | /api/tenants/:tenantId/assets | List assets |
 | POST | /api/tenants/:tenantId/assets | Create an asset |
+| GET | /api/tenants/:tenantId/assets/:assetId | Get asset metadata |
 | POST | /api/sidecars | Register or update a sidecar |
 | GET | /api/sidecars | List all sidecars |
 | GET | /api/sidecars/:id | Get a sidecar by ID |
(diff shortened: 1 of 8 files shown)


alexanderguy (commented):
Commit-message style audit clean — subjects and body lines under 72, no prefixes, no tracker IDs, no prior-commit references, no filenames, no trailing punctuation. Commit-message coherence verified against the diffs. `make all` green at HEAD.

PR #56, opened 2026-06-09
Distribute NPM-compatible tool packages end-to-end

alexanderguy:
Implements [INTR-108](https://linear.app/abklabs/issue/INTR-108/distribute-npm-compatible-tool-packages-end-to-end).

Agents declare their tools via npm-style pins in `agent.tool_packages`. The hub resolves each pin's transitive closure against configured registries (HTTP-backed via `npm-registry-fetch` or asset-backed via the `package-registry` asset kind), writes the resolved manifest into the deploy pack, and ships it to the sidecar. The sidecar materializes every entry through a content-addressable tarball cache, builds a per-instance nested `node_modules/` layout (hardlinks + symlinks) so each top-level package's `interchange.tools` entry can `require()` / `import` against its closure without host help, dynamic-imports each top-level, and atomically swaps the new bundle into place. Rejected applies route through the `deploy.apply.error` frame channel and the agent's audit trail.

Three commits:

1. `Add the tool-packaging subsystem` — closure resolver, sidecar loader, content-addressable cache, atomic-apply protocol, manifest/pin schemas, plugin-factory infrastructure in `@intx/agent`, packable bundles for the posix/mail/lsp tools.
2. `Replace the harness wiring with the loader-driven pipeline` — default harness rewritten around the loader, sidecar boot moved into a small config module, `pack-receive` gains hooks for asset-backed tarballs.
3. `Wire the tool-package manifest through the deploy pipeline` — session service resolves pins per agent at session launch, asset service learns the `package-registry` kind, hub-agent deploy-tree carries the manifest, REST routes for tarball upload/list/delete, dev orchestrator publishes the workspace builtins and seeds example agents.

## Test plan

- [x] `make all` green at HEAD (3058 tests pass, 0 fail)
- [x] `make all` green at every commit on the branch (verified individually)
- [x] Commit-message style audits empty (`awk 'length > 72'` on `%s` and `%b`; grep for prefix patterns, tracker IDs, prior-commit references, filename refs in subjects)
- [x] Branch's final tree differs from the pre-reshape source on five files only; each difference is a quality fix (real `tar.TarOptionsWithAliasesAsyncFile` instead of phantom `tar.CreateOptions`, `@types/ssri` instead of hand-rolled `declare module`)


alexanderguy (commented):
Three commits, 98 files, +24015/-289. No `Bin` markers.

Commit-message style: `awk 'length > 72'` empty on `%s` and `%b`; grep for prefix patterns, tracker IDs, prior-commit references, and filename refs in subjects all empty. Coherence: each `git show` matches its message; commit 2's body names the deploy-flow test deletion it triggers.

`make all` exits 0 at each commit's SHA (2985 / 3058 / 3059 tests).

Read every source file in `@intx/tool-packaging` end-to-end (9,273 lines including tests). Looked specifically for: path-traversal containment at tarball-entry, asset-mount, and symlink boundaries; integrity verify-before-write and verify-before-extract; atomic-apply rollback paths and the diverged-disk throw; LRU eviction with refcount-deferred reclaim; resolver subtree poisoning and queue dedup. No issues found.

Ready to ship.

PR #57, opened 2026-06-10
Add the workflow framework and per-call authz context

alexanderguy:
## Summary

- Add the @intx/workflow package: workflow-run state machine, definition surface, single runtime body, and runLocal in-memory env.
- Widen the authz callback with a generic per-call context the workflow runtime supplies as `{ stepId, attempt, runId }`.

## Verification

- `make all` is green: the main suite and the second suite both report 0 failures.
- runLocal exercises step, map, gate, awaitSignal, sleep, childWorkflow, and escalation end-to-end against the dispatch-shape fixture.
- The runtime body has no environment-shaped discriminator; an explicit source-level test enforces the discipline.
- Resume from a partial log rehydrates step outputs through the blob substrate and runs to completion.
- AuthorizeContext propagation: an integration test asserts authz calls inside a workflow step receive `{ stepId, attempt, runId }` while bare callers see `{}`.

Closes INTR-155
Closes INTR-168


alexanderguy:

Whole-branch review (`base..HEAD`) returned clean. Evaluated:

- Commit-message audits against `style`'s subject/body rules and the project's existing log convention — five commits, no prefixed subjects, bodies carry motivation rather than diff recap or session narration.
- Diff scope against the issue boundaries (INTR-155 surface + state machine + `runLocal`, INTR-168 `AuthorizeContext`) — every changed file is part of `packages/workflow` or its tests; no drive-bys.
- Event vocabulary, transition function, and resume contract in `state-machine/` against §4 of the workflow framework spec — cancellation, signal, and timer invariants enforced; `consumedMessageId` dedup committed to the run-scoped set; `CancelOrigin` shape-validated at the state layer with principal validation deferred to the `workflow-run` kind handler.
- `WorkflowRuntimeEnv` against the single-runtime-body discipline — the source-level grep in `run.test.ts` keeps `isChildProcess` / `processRole` / `runtimeMode` out of the body; env keys carry every per-host difference.
- `runLocal` against the env contract — `ephemeral: true` on the in-memory blob substrate, in-process scheduler, in-process signal channel; the runtime body sees only the interface.
- Primitive runners against the spec's per-primitive semantics — `awaitSignal` commits paired `TimerSet` + `TimerFired` around the timeout, `childWorkflow` propagates failed/cancelled child as `StepFailed` on the parent's spawn step, gate skip closure preserves diamond joins.
- `AuthorizeContext` against the `@intx/agent` authorize surface — generic widens with per-call `{ stepId?, attempt?, runId? }`, callers in workflow runtime supply the context at every `env.authorize` call.

PR #59, opened 2026-06-15
Workflow asset/deploy, authz, sidecar runtime, and trivial-workflow uniformity

alexanderguy:
## Summary

Lands the workflow runtime end-to-end: the `@intx/workflow` definition surface and state machine, the `@intx/workflow-host` per-deployment supervisor + child IPC, the `@intx/workflow-deploy` orchestrator with capability walk and operator-approval gate, the hub-side `workflow` (asset) and `workflow-run` (event log + claim-check substrate) kind handlers, and the sidecar wiring that branches between the trivial and multi-step deploy paths. Trivial-workflow uniformity holds: the existing hub agent-deploy integration tests run against the workflow-deploy orchestrator with their assertions bit-preserved.

## What ships

**Workflow definition + runtime.** Workflow definition primitives (`step`, `gate`, `map`, `sleep`, `awaitSignal`, `childWorkflow`); a typed state machine with phase transitions, terminal-phase locks, and sequence monotonicity; a runtime body that drives the DAG against a `WorkflowRuntimeEnv` and is testable in-process via `runLocal`. The runtime body admits cancel across the early-lifecycle window and the post-loop terminal commit; selector paths surface missing keys and out-of-range indices loudly; mid-step resume surfaces an explicit `RuntimeResumeUnsupportedError` rather than stalling.

**Workflow-host supervisor + child IPC.** One per-deployment supervisor owns a `Bun.spawn`'d workflow-process child connected over an Ed25519-signed control channel and an HMAC-secured event channel. The supervisor drives a FIFO mail dispatch loop with claim-check substrate primitives (inbox → processing → terminal → consumed). Spawn-time `replayProcessingToInbox` runs before the dispatch loop's first dequeue; the recycle path drives drain → replay → cohort abort → kill → respawn against a live cohort so drain accumulators see real terminal events; `shutdownInternal` awaits the replay and the in-flight dispatch loop. Pending merge round-trips and `markConsumed` waiters that the supervisor accumulates from `substrate.write.request` IPC reject through `MergeAbortedError` on cohort transition so handler closures never outlive the dying cohort. Terminal events route through the spawning cohort's broadcaster captured at pump-start time. Substrate writes from the child proxy through a `substrate.write.request` IPC frame so the supervisor remains the sole writer to the workflow-run repo. `installNewChild`'s phase guard kills the freshly spawned child and releases its event-channel resources when a shutdown lands during the kill/respawn gap. The public `drain()` admits `running` and `starting` only; the recycle path's internal drain bypasses through `drainImpl({}, { fromRecycle: true })` so the live controlSender receives the kill-prep frame without exposing external callers to the silent-loss window.

**Workflow deploy.** Orchestrator that materializes the workflow asset, walks capability declarations to emit `tool:*`, `capability:*`, `inference.source:*`, `director:*`, `mail:*` grants, and gates the operator approval set against every grant the walk emits. The trivial-wrap path projects `HarnessConfig.tools` onto walk-visible factories so the approval gate enforces real tool grants on the legacy agent-deploy path. The multi-step branch pins every step's inference source ahead of provisioning so an unapproved source fails the whole deploy before `launchSession` provisions an agent-state repo at the sidecar.

**Hub-sessions kind handlers.** `workflow` (asset) validates envelope shape, capability declarations, and tool/director ref grants; rejects array-shaped `steps` and `state`. `workflow-run` enforces append-only on events, blobs, and the address/consumed index; rejects deletions of immutable prior entries (events, blobs, consumed, processing, inbox); enforces sequence contiguity from seq 0; scopes `workflow-process` writes by `principal.runId` so a child for run R1 cannot write into R2's namespace; fails closed on malformed `workflow-process` principals at the path-scope check so a substrate wired with a permissive authorize callback cannot bypass enforcement. The substrate's per-commit walk in `receivePack` builds full-view closures from the prospective commit's tree, so the kind handler's prior-walk sees a true representation of what the writer will land. `createAsset` rejects every repo kind the asset service does not back (`agent-state`, `workflow`, `workflow-run`) so the write-side surface matches the read-side `rowToAsset` narrow.

**Hub-agent + sidecar wiring.** Sidecar deploy router branches on the multi-step deploy frame; per-(repoId, ref) pack-push queue with bootstrap-retry coalescing; signal / drain / mail inbound routers with symmetric `unregister` APIs; `agent.undeploy` shuts down the per-deployment supervisor, unregisters every router, prunes the bootstrap-bookkeeping set, and rolls back failed deploys cleanly. The router tracks slug claims so two agent addresses that project to the same `deriveTrivialDeploymentId` slug surface a structured collision at the second deploy instead of silently overwriting the first deploy's workflow-run repo. Both deploy branches wrap their bodies so a failure between `claimSlug` and `registerDeployment` releases the slug; the multi-step branch additionally unwinds its supervisor (killing the workflow-process child), its `activeSupervisors` entry, and its three router registrations so a `registerDeployment` failure does not orphan the freshly spawned cohort.

**Integration tests.** End-to-end matrix covering trivial-roundtrip (the uniformity claim), multistep-signal, drain-roundtrip (cancel-mode and wait-mode companions), child-workflow-roundtrip (including grandchild recursion and 5-siblings; parent-namespace isolation positively asserted), unresolvable-director, FIFO mail (3-mail correctness + 15-mail load), mail-edge-cases (no-Message-Id / malformed / duplicate, with log-substring assertions on the supervisor's `claim_check_already_*` rejection path), and pending-merges-cohort-abort. The deploy-flow integration test runs against the workflow-deploy orchestrator with its assertions bit-preserved.

## Validation

- `make build`: clean (`tsc -b --noEmit` exits 0)
- `make lint`: clean (prettier + eslint + `bun bin/gen-api-docs.ts --check`)
- `make test`: clean — fast pass (~40s, 3336 tests) plus slow integration pass (~120s, 308 tests / 28 skip at `--timeout 120000`); stable across consecutive `make all` runs
- `bun run test:load`: clean — 15-mail FIFO load case (~70s) held out of `make test` per INTR-194's O(N²) `validatePush` cost

## Test-script topology

`test` runs two `bun test` passes. The first enumerates unit-test directories at the default 5s timeout. The second enumerates integration files explicitly with `--timeout 120000`; the 120s per-test cap absorbs the 5-siblings child-workflow scenario's worst-case wall clock under parallel test load. The 15-mail FIFO load case lives in `fifo-mail-load.test.ts` and runs through a dedicated `test:load` script (CI invokes that separately) — its per-iteration wall clock is bounded by `validatePush`'s O(N²) byte-equality check on the receive side, tracked as INTR-194.

`bunfig.toml` carries only the `[test]` header; bun 1.2.22's `[test].exclude` and `pathIgnorePatterns` are both non-functional for this discovery shape, so directory selection happens through the positional arguments in the `test` script. The integration suites land in the second pass; CI sees both pass summaries.

## Validation criteria (INTR-156 issue body)

- [x] `workflow` and `workflow-run` kind handlers ship; unit tests cover §4 invariants
- [x] Capability-walk tests for trivial and multi-step workflows produce expected grant unions
- [x] Supervisor + child-process runtime tests: drain ✓, recycle ✓, crash-restart (deferred — see follow-up), self-discovery (deferred — same follow-up)
- [x] Mail-trigger serialization test (15-mail FIFO under load passes; 50-mail target deferred — see follow-up)
- [x] Trivial-workflow round-trip end-to-end test passes
- [x] Multi-step round-trip end-to-end test passes
- [x] Drain round-trip end-to-end test passes
- [x] Existing hub agent-deploy integration tests pass against the new workflow-deploy path without test-assertion modification — trivial-uniformity claim verified
- [x] `make all` is green

## Follow-up issues

- **INTR-193** — Auto-respawn workflow-host child on unexpected exit, with crash-loop guard. Owns the crash-restart and self-discovery validation criteria from this issue.
- **INTR-194** — OID-equality short-circuit on `validatePush` byte-equality checks. The substrate-level optimization needed to make the 50-mail FIFO load test CI-tractable.


alexanderguy:
# Code review

## Reviewer-of-record checks

Run against the current branch state with base `main`.

- `git diff main...HEAD --stat` — 185 files, +54,290 / −1,135. Scanned for `Bin` markers on source / markdown / config files: none. Scanned for files outside `packages/`, `apps/`, `tests/`, `docs/`, `examples/`, `bin/`, `tsconfig*`, `eslint.config.ts`, `CONVENTIONS.md`, `Makefile`, `package.json`, `bunfig.toml`, `bun.lock`: none.
- `git log --oneline main..HEAD | wc -l` — 116 commits on branch; scope matches the INTR-156 issue body. Spot-checked commit messages against `git show <sha>` for a sample across the runtime, substrate, supervisor, and integration-test areas; diffs match messages.
- Subject style (`git log main..HEAD --format='%s' | awk 'length > 72'`): empty. Prefix-style subjects (`feat:`, `fix:`, `[tag]`, ticket IDs, status tags): grep matched none. Filename references in subjects: none. Trailing punctuation: none.
- Body style (`git log main..HEAD --format='%b' | awk 'length > 72'`): empty. External tracker references (`INTR-`, `JIRA-`, `Closes/Fixes/Refs`, `#NNN`): grep matched none. SHA cross-references in bodies: none. Commit-series cross-references ("prior commit", "earlier commit", "next commit"): grep matched none.

## Substrate layer — `packages/hub-sessions`

Verified the `workflow-run` invariants by direct read + test citation:

- Per-commit walk during `receivePack` rejects `pack_walk_dangling_parent` for the workflow-run kind in `store.ts`; pinned by the multi-commit pack-walk test in `store.test.ts` and by `workflow-run-kind.test.ts`'s single-pack-validation case (the test composes the source repo's genesis, enqueue, and dequeue commits' reachable OIDs into one pack so the receiver's per-commit walker sees every parent it needs).
- Append-only across `events/`, `blobs/`, `consumed/`, and `processing/` with absence + byte-divergence rejection in `workflow-run-kind.ts`. Six deletion-bypass cases sit in `workflow-run-deletion-bypass.test.ts` covering event-blob drop, immutable-blob drop, consumed-dedup drop, processing-drop without consumed, addresses-wipe with prior consumed, and runs-wipe with prior terminal.
- Principal-vs-origin enforcement at `validatePush` via `CANCEL_ORIGIN_TO_PRINCIPAL_KIND`. Hub-admin requires a hub principal; self, supervisor-drain, and supervisor-operator require a supervisor principal.
- Workflow-process path scope: `enforceWorkflowProcessPathScope` rejects every `workflow-process` write under `addresses/` and any `runs/<X>/` write whose `runId` segment does not match the principal's `runId`. Hub and supervisor principals retain broader scope by design.
- `enforceWorkflowProcessPathScope` fails closed on malformed `workflow-process` principals so a substrate wired with a permissive authorize callback cannot bypass enforcement. Pinned by the `workflow-process path-scope fail-closed` case in `workflow-run-kind.test.ts`.
- Sequence contiguity: `enumerateRunEvents` verifies `entries[i].filenameSeq === baseSeq + i` so a missing seq surfaces as `sequence_gap` with the offending runId and seq.
- Closure-honesty fix at `writeTreeUnderLock` unions `content.files` with surviving prior-tree paths so the kind handler's prior-walk sees a true representation of what the writer will land.
- `workflow-kind.ts` envelope schema declares `steps` and the optional `state` as record-rejecting-arrays so an array-shaped `steps` is rejected at the substrate edge rather than reaching downstream consumers as an array.
- `createAsset` rejects every repo kind the asset service does not back (`agent-state`, `workflow`, `workflow-run`) so the write-side surface matches the read-side `rowToAsset` narrow.

## Supervisor cohort lifecycle — `packages/workflow-host/src/supervisor`

Traced phase transitions against the source and verified each is pinned by a regression test:

- `spawn → ready → running` including the "control channel ends before ready" path; pinned by `lifecycle-races.test.ts`.
- Cohort awaiter rejection on recycle: `MergeAbortedError` via `rejectCohortAwaiters` fires before the new dispatch loop starts. Pinned by `substrate-write.test.ts`'s cohort-abort cases.
- Late terminal-event from a prior cohort routes to the prior broadcaster, not the new one — closures over `cohortBroadcaster` are captured at pump-start time; disposed broadcasters no-op. Pinned by `stale-cohort-routing.test.ts`.
- `deliverSignal` rejects during `recycling`; pinned by `recycle.test.ts`.
- `recyclePolicy.stop()` and the re-arm prevention via `if (stopped) return`.
- Unsubscribe-during-emit safe via snapshot copy in `terminal-broadcaster.ts`.
- Shutdown-during-recycle race: `installNewChild`'s phase guard kills the orphan child and releases its `eventPump` and `controlIncoming` resources with structured warns. Pinned by `recycle.test.ts`'s "shutdown during the kill/respawn gap" case; the test's racing precondition is pinned by a `setImmediate` yield before the second-child ready is driven.
- External `drain()` during `recycling` is a silent no-op. The recycle path's internal call routes through `drainImpl({}, { fromRecycle: true })` to bypass the rejection. Pinned by the `external drain phase guard` case. The drain-frame counter uses a typed `SignedEnvelope` + `ControlPayload` parse so a frame whose serialized text contains the substring `"drain"` cannot inflate the count.

## Runtime body — `packages/workflow/src/runtime`

- `executeRunBody` admits a racing cancel against both the initial `RunStarted` commit and the post-loop terminal commit via `settleCancelling`, so a `cancel()` landing between the last in-loop reload and the terminal commit settles as cancelled rather than rejecting with `TransitionError(code="phase")`. Pinned by `cancel-early-lifecycle.test.ts` and `cancel-during-resume.test.ts`.
- `resolvePath` rejects both missing keys (`!(segment.key in cursor)`) and out-of-range indices (`segment.index < 0 || >= cursor.length`) with `SelectorError`. Pinned by `selectors.test.ts` and `selectors-edge.test.ts`.
- `runStep` canonicalizes `rawInput === undefined ? null : rawInput` once before both `recordOutput` and `invokeStep` see the value, so the audit blob and the agent observe the same value. Pinned by `audit-input-divergence.test.ts`.
- Mid-step resume against a seed log whose tail is `SignalAwaited`, `TimerSet`, or a mid-`map` state raises `RuntimeResumeUnsupportedError` from `executeRunBody` after the seed-log replay. The class names the step and its awaited primitive. Pinned by `resume-mid-map.test.ts` and `resume-awaiting-signal.test.ts`.
- `runtimeRun`'s docstring describes the contract honestly: complete-or-cancelled or boundary-aligned seed logs resume; mid-step resume is unsupported and surfaces as `RuntimeResumeUnsupportedError`.

## Sidecar wiring — `apps/sidecar/src/workflow-host-wiring.ts`

- `validateWorkflowProjection` enforces the four invariants the wire shape leaves open (`definition.id` non-empty, `stepOrder` non-empty, `STEP_ID_PATTERN`, `steps[id]` presence). The `sources[id]` check duplicates the arktype narrow deliberately to surface a structured router-side error.
- `deriveTrivialDeploymentId` slug-collision rejection via the shared `slugClaims` map. Pinned by `workflow-host-wiring.test.ts: deriveTrivialDeploymentId slugs collide`. Same-address re-claim is a no-op; the slug becomes claimable again after `undeploy`.
- Deploy-failure unwind on both branches:
  - Trivial branch: `try { ... } finally { releaseSlug if !claimedSlugSucceeded }`. Pinned by the failed-then-retry test in the trivial-frame regression describe.
  - Multi-step branch: tracks `wiredForUnwind`, `supervisorRegistered`, `routersRegistered` and unwinds in reverse-registration order (routers → `activeSupervisors` entry → `supervisor.shutdown()` → slug release) so a `registerDeployment` failure does not orphan the freshly spawned workflow-process child or leak the router registrations. Pinned by `workflow-host-wiring-deploy-failure.test.ts`.
- `agent.undeploy` awaits the per-deployment supervisor's `shutdown()` so the workflow-process child is not orphaned across undeploy. Pinned by `workflow-host-wiring-undeploy-supervisor.test.ts`.
- `registerDeployment` runs last on both deploy branches — only after every throwy operation succeeds — so a mid-deploy failure cannot leave the registry mapping a deploymentId to an agent address that never finished provisioning. Pinned by `workflow-host-wiring-deploy-failure.test.ts`.

## Hub-agent — `packages/hub-agent/src/ws/hub-link.ts`

- `mail.inbound` arm wraps `tryRoute` in `try { ... } catch { ... }` matching the `signal.deliver` and `drain.deliver` arms, and the `handleMessage` chain grows a tail `.catch(...)` so any future arm that throws degrades to a dropped frame rather than wedging the link. Pinned by `hub-link-mail-router.test.ts`.
- `handleAgentUndeploy` calls `deployRouter.undeploy?(...)` before `sessions.destroySession` so router state for the undeployed agent is released before a racing inbound frame can reach the now-torn-down supervisor. Pinned by `workflow-run-pack-client-undeploy.test.ts`.
- `handleAgentUndeploy` also prunes `workflowRunPackBootstrapped` and `workflowRunPackBootstrappedByAddress` for the undeployed agent so the bootstrap flag set does not grow without bound and a redeploy after a hub-side workflow-run reset does not skip the bootstrap retry. Pinned by `hub-link-bootstrap-prune.test.ts`.

## Integration tests

- `deploy-flow.test.ts` codifies the lifecycle preconditions (launchSession + the first message routing) in `beforeAll` so each individual test in the chain passes both end-to-end and under a `bun test -t <name>` filter. The `beforeAll` waits for `message.run.ended` after routing the setup message so the downstream sync test reads populated agent state from disk.
- `fifo-mail.test.ts` budgets 90s for `waitForConsumedEntries`. The supervisor's per-message `markConsumed` push traverses a hub-ack-await for every prior push in the dispatch loop, so the burst's final `markConsumed` lands well after the third run's terminal event observation; the budget sits under the 120s bun per-test cap.
- Slow integration pass runs at `bun test --timeout 120000` so the 5-siblings child-workflow scenario fits inside the per-test cap on parallel-loaded machines.
- `workflow-run-kind.test.ts` single-pack-validation case composes the source repo's genesis, enqueue, and dequeue commits' reachable OIDs into one pack so the receiver's per-commit walker sees every parent it needs and does not hit `pack_walk_dangling_parent`.

## Tests

- `make build`: `tsc -b --noEmit` exits 0.
- `make lint`: prettier + eslint + `bun bin/gen-api-docs.ts --check` all clean.
- `make test`: fast pass 3336 / 0 fail; slow pass 308 / 28 skip / 0 fail.

## Convention compliance

- Zero `TODO` / `FIXME` / `XXX` markers in the branch's net diff (`git diff main...HEAD | grep -E "^\+.*\b(TODO|FIXME|XXX)\b"` empty).
- Every `as` cast in production paths I sampled carries an `eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion` with a justification.
- Timeouts named via constants (`DEFAULT_DRAIN_TIMEOUT_MS`, `DEFAULT_KILL_TIMEOUT_MS`, `DEFAULT_POLICY_INTERVAL_MS`, `DEFAULT_TERMINAL_WRITE_WATCHDOG_MS`, `MAX_BUFFERED_MAIL`); no magic numbers in new code.
- No filename:line references in code comments that would rot under future edits.

## Defensive-coding surfaces

- `replayProcessingToInbox` failure on spawn is documented best-effort: the dispatch loop blocks on `replayDone` via `Promise.race` and the `.then(wakeDispatch)` chain fires only on success. The failure surfaces via `logger.warn`.
- `handleSubstrateWriteRequest` drops `substrate.write.request` frames received outside the running phase with a `logger.warn`; the docblock names why this is the only viable action (no sender to respond on; the child's read pipe is being torn down alongside the transition).
- The trivial deploy's `driveTrivialRunChain` audit chain is best-effort against the deploy path; the warn surfaces persistent failures without killing the agent's reactor.
- The drain-timeout terminal-event watcher's iterator-error path warns and falls through to the outer `setTimer` deadline; the inline comment in `drain-timeout.ts` names the contract.
- The recycle phase guard's `eventPump` and `controlIncoming` cleanups warn-log instead of swallowing errors, so a broken-pipe condition on an orphan cohort reaches operator logs.

PR #61, opened 2026-06-18
Carry multimedia attachments from the hub to model adapters

alexanderguy:
## Summary

- The mail route accepts base64-encoded attachment bytes, validates each
  against a MIME allowlist and per-attachment/per-message size limits, and
  forwards the decoded bytes to the agent. Rejections return ordered,
  per-index structured error codes.
- Conversation messages carry attachments inside the signed PGP/MIME
  envelope as a `multipart/mixed` body; received attachments are parsed
  back and mapped to the model content blocks each provider adapter
  marshals (image/video/audio by major type, documents by allowlist).
- An unmappable inbound attachment type degrades to a visible text marker
  rather than failing turn construction.

## Verification

- `make all` passes: lint, `tsc -b`, and tests (3386 + 313 tests, 0
  failures).
- An end-to-end test posts an image attachment, delivers it through the
  real transport `fetchFull`, builds the turn, and marshals it with each
  shipped adapter (Anthropic, OpenAI, Google), asserting the original
  bytes reach every provider's wire format.

Closes INTR-195

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -651,9 +651,10 @@ Persists the user message as a mail record and dispatches it to the running agen
 Body: SendMessage
 
 201: MailResponse -- Mail sent
-400: ErrorResponse -- Validation error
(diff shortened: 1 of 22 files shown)


alexanderguy:

PR #62, opened 2026-06-19
Make workspace dependency declarations honest, converged, and guarded

alexanderguy:
## Summary

- Every workspace package declares the modules it imports — including the
  type-only and outside-`src/` imports that previously resolved only through
  Bun's flat `node_modules`.
- Every external dependency used by two or more packages is referenced as
  `catalog:`, its version defined once in the root catalog (eight new entries
  join `arktype`).
- `bin/check-deps.ts`, wired into the `lint` chain (and so the pre-commit
  hook), fails the build on an undeclared import, a shared dependency not on
  the catalog, a catalog entry with fewer than two consumers, or lockfile
  drift. It also checks `bin/` and `tests/` against the root manifest (which
  declares the externals they use), and its `checkWorkspace()` core is covered
  by `bin/check-deps.test.ts` against synthetic workspace fixtures.
- `bun.lock` matches the committed manifests.

## Verification

- `bun run check` (tsc), `bun run lint` (prettier, eslint, docs, and
  `check:deps`), and `bun test` (3413 + 318 tests, 0 failures) all pass.
- `bun install` reports no version changes — the catalog migration and the
  newly declared `bin/`/`tests/` externals are resolution-neutral.
- `bun run check:deps` passes on the branch and is confirmed by the fixture
  tests to flag a re-introduced undeclared import (including inline
  `{ type X }` and `typeof import()` forms), an off-catalog shared
  dependency, a single-consumer catalog entry, an undeclared `bin/`/`tests/`
  import, and lockfile drift.

Closes INTR-205


alexanderguy:
Self-review with the `code-review` skill. `bun run check`, `bun run lint`
(including `check:deps`), and the full `bun test` suite pass; the guard's
checks are mutation-verified to fire, and `bin/check-deps.test.ts` covers
its `checkWorkspace()` core against synthetic workspace fixtures.

Two intentional limitations remain, both documented in the script header:
- Triple-slash `/// <reference types="...">` directives are not extracted.
- The lockfile check uses `bun install --frozen-lockfile`, which validates
  dependency resolution but not every lockfile bookkeeping detail (e.g.
  bin-map or workspace-link metadata).

PR #63, opened 2026-06-19
Add a tenant-scoped model catalog with ordered inference-source resolution

alexanderguy:
## Summary

- Add a tenant-scoped catalog of models, model providers, model×provider
  offerings, and append-only multi-currency pricing, inherited through the
  tenant hierarchy with credentials-style add / disable / shadow semantics.
- Resolve an agent's model requirements against three sources — tenant
  catalog, creator definition, and invoker launch preference — into an
  ordered `InferenceSource[]` keyed by catalog primary keys, replacing the
  old `${provider}:${model}` synthesis; the runtime fails over across the
  list on source-specific errors and resets to the preferred source each
  inference cycle.
- Expose full CRUD + a tenant model-discovery endpoint, an admin UI for the
  four catalog entities, seed data, and updated architecture and package
  docs.

## Verification

- `make all` passes: lint, `tsc -b --noEmit`, and the full suite (3499 + 344
  tests, 0 failures).
- `docs/API.md` regenerates clean and reflects the new catalog routes.

Closes INTR-196


alexanderguy:
Self-review (`code-review` skill) returned clean. By "clean" I mean the branch holds on each dimension the review covers:

- **Tenant isolation.** Every catalog read and write derives the tenant from the authenticated context, never the request body; mutations act only on rows the tenant owns, and an edit to a non-owned or inherited row returns 404. No path reads or writes across tenants.
- **Boundary validation.** Malformed input is rejected at the edge with a clean status rather than deep in the stack: exactly one of credential/wallet on a provider (empty strings normalized to absent), one requirement per model on a definition, a parseable `effectiveFrom`, and a credential that actually resolves within the tenant's ancestor chain.
- **Resolution correctness.** Inheritance resolves by `(canonical model, provider name)` with nearest-tenant-wins shadowing and disable-suppression; distinct model requirements map to distinct offerings, so the ordered `InferenceSource[]` carries no duplicate ids.
- **Failure handling.** Errors surface rather than being swallowed; the one fire-and-forget path (the post-mutation source push) is wrapped so a failure logs instead of becoming an unhandled rejection.
- **Failover behavior.** The reactor separates source-specific failures (fail over), rate limits (short same-source wait, then fail over), and source-invariant failures (abort), and restarts at the preferred source each cycle.
- **Coverage.** The new resolution, pricing selection, failover, and route validation branches are exercised by tests; `make all` is green and `docs/API.md` regenerates with no drift.

PR #64, opened 2026-06-22
Fix from-scratch bring-up on main: make build and asset seed 409

alexanderguy:
## Summary

- Restores a clean `make build` on `main`. An inference rename from a single `source` to an ordered `sources` list plus `defaultSource`, and a new `setSources` agent method, had left three consumers behind, so `tsc -b` failed.
- Fixes the documented quick-start `bin/db-reset && bun bin/dev.ts --seed`, which aborted with HTTP 500 during seeding. Duplicate asset inserts now map to 409 because the Postgres SQLSTATE is read from the Drizzle-wrapped error's `cause` chain instead of only the top-level error.

## Changes

- `examples/posix-demo/src/cli.ts`, `examples/ring-demo/src/cli.ts`, `packages/workflow-host/src/adapters/step-invoker.test.ts`: migrate each call site to `sources: [source]` + `defaultSource: source.id`, and add a `setSources` stub alongside the existing `setSource`. Test and example code only.
- `packages/db/src/pg-error.ts` (new): `pgErrorCode` walks the error `cause` chain (depth-bounded, reuses `hasCode`) and returns the first SQLSTATE found, plus the `PG_UNIQUE_VIOLATION` / `PG_FOREIGN_KEY_VIOLATION` constants. Lives in `@intx/db`, the package that owns Drizzle and the postgres driver.
- `packages/hub-sessions/src/asset-service.ts`: the three constraint-violation guards use `pgErrorCode`, keeping each `throw err` fall-through so unrecognized errors still surface.
- `packages/hub-sessions/src/asset-service.test.ts`, `packages/hub-api/src/routes/assets.test.ts`: fixtures nest the SQLSTATE under `cause` to mirror how Drizzle throws, with a regression test for the wrapped shape.

The seed itself is unchanged: its conflict-tolerant single POST is correct once the hub returns 409.

## Testing

- `make all` green: 3521 unit tests and 344 integration tests pass, 0 fail.
- Mutation check: reverting the guard to the old top-level-only check fails 4 tests, confirming the regression coverage is real.
- Live end-to-end: `bin/db-reset && bun bin/dev.ts --seed` completes a full seed; the `workspace-builtins` re-create returns 409 instead of 500.

Closes INTR-213

### DIFF
--- a/examples/posix-demo/src/cli.ts
+++ b/examples/posix-demo/src/cli.ts
@@ -335,7 +335,8 @@ const betaDef = defineAgent({
 });
 
 const alphaEnv: MailEnv = {
-  source: alphaSource,
+  sources: [alphaSource],
+  defaultSource: alphaSource.id,
   storage: storageAlpha,
   workdir: alphaDir,
   audit: noopAuditStore(),
@@ -346,7 +347,8 @@ const alphaEnv: MailEnv = {
 };
 
 const betaEnv: MailEnv = {
-  source: betaSource,
+  sources: [betaSource],
+  defaultSource: betaSource.id,
   storage: storageBeta,
   workdir: betaDir,
   audit: noopAuditStore(),
--- a/examples/ring-demo/src/cli.ts
+++ b/examples/ring-demo/src/cli.ts
@@ -394,7 +394,8 @@ for (const [i, name] of AGENT_NAMES.entries()) {
   });
 
   const env: MailEnv = {
-    source,
+    sources: [source],
+    defaultSource: source.id,
     storage,
     workdir: join(TMP_ROOT, name),
     audit: noopAuditStore(),
--- a/packages/db/src/index.ts
+++ b/packages/db/src/index.ts
@@ -1,4 +1,9 @@
 export { createDB, type DB } from "./client";
+export {
+  pgErrorCode,
+  PG_UNIQUE_VIOLATION,
+  PG_FOREIGN_KEY_VIOLATION,
+} from "./pg-error";
 export type { DBConfig } from "./config";
 export { runMigrations, dropSchema } from "./migrate";
 export { createGrantStore } from "./grant-store";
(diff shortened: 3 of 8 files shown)


alexanderguy:
## Self-Review Summary

Reviewer-of-record pre-PR self-review of the full `origin/main..HEAD` range (3 commits), all checks run in-session against the worktree.

### Verification (cited)
- **Full gate:** `make all` exited clean — 3521 unit + 344 integration tests pass, 0 fail.
- **Mutation check:** reverting the asset-service guard to the old top-level-only SQLSTATE check fails 4 tests (incl. the new wrapped-cause cases), proving the regression coverage is real.
- **Live end-to-end:** `bin/db-reset && bun bin/dev.ts --seed` completes a full seed against a fresh database; the `workspace-builtins` re-create returns 409 (previously a fatal 500).
- **Commit-message audit:** subjects ≤72, no prefixes/paths/trailing punctuation; no body lines >72; no tracker/cross-commit/PR references.
- **Scope/binary:** `git diff --stat` shows 8 source/test files, no binary markers, nothing out of scope.
- **Docs:** `make lint` runs the API-docs freshness check; `docs/API.md` is up to date.

### Files Reviewed
- `examples/posix-demo/src/cli.ts`, `examples/ring-demo/src/cli.ts`: env migrated to `sources` + `defaultSource`. Mechanical, no behavior change.
- `packages/workflow-host/src/adapters/step-invoker.test.ts`: env migrated; `setSources` stub added alongside `setSource`.
- `packages/db/src/pg-error.ts` + `packages/db/src/index.ts`: new `pgErrorCode` cause-walk + SQLSTATE constants, exported from `@intx/db`.
- `packages/hub-sessions/src/asset-service.ts`: three guards switched to `pgErrorCode`; `throw err` fall-throughs preserved.
- `packages/hub-sessions/src/asset-service.test.ts`, `packages/hub-api/src/routes/assets.test.ts`: fixtures nest SQLSTATE under `cause`; regression test added.

### Issues Found and Resolved
- A critique pass on the asset-service change found no blockers/no should-fix items. Its one nit — the hub-api route-test fixtures still fabricated the SQLSTATE at the top level and no longer mirrored how Drizzle throws — was fixed on this branch (the third commit) rather than deferred.

### Scope note
The separate process gap (no automated typecheck gate let the broken `make build` reach `main`) is intentionally out of scope here and tracked under the "Interchange CI chain" project.

PR #65, opened 2026-06-22
Run multi-agent workflows on a unified execution host

alexanderguy:
## What this does

- **Unified execution host.** The sidecar spawns a workflow-process child that runs a real agent for each step — materializing its tools, threading agent events to the hub timeline, keeping the agent warm across messages, and persisting its conversation durably across a child respawn. The workflow runtime now does real agent work (previously a stub step-invoker).
- **INTR-209 — launchable workflow.** A seeded 3-step human-in-the-loop workflow (`step → awaitSignal(approve) → step`) is launchable and approvable from the admin UI, with run-trigger and observe routes plus UI.
- **INTR-208 — convergence.** Agent-launch and workflow-launch deploy through one path; the trivial single-step branch is removed.

## Durable-substrate latency hardening

Validating the unified path surfaced that per-message latency *grew* over a deployment's lifetime — the git-backed run substrate did more work per commit as it accumulated runs and dedup entries. Root-caused by measurement to per-commit O(N) costs, and fixed:

- **D1** — the warm-agent conversation is an O(1) append (a write-ahead log plus periodic checkpoint), not a whole-blob rewrite.
- **B2** — a run's events batch into one durable commit per execution segment (was four), with crash recovery preserved.
- **B3a** — the two largest per-commit walks (the index status walk and the run validation walk) are scoped to the prefix the commit actually touches.
- **P2** — the dedup index is bounded by a configurable retention watermark (24h default), with exactly-once preserved and no gap.
- A per-run local-disk scratch leak is reclaimed: cold/multi-step on run completion, warm on undeploy, the durable conversation store untouched.

Three of the four O(N) per-commit costs are now flat or bounded; the unified path no longer meaningfully degrades over a deployment's life. The smallest residual lever (a per-ref index for `resetIndexToRef`) is deferred as INTR-214. The receive-side `validatePush` enumeration that previously capped sustained mail throughput (INTR-194) is resolved by B3a's scoping.

## Why it is sound

- `make all` is green on HEAD: lint, build, and 364 integration tests plus 3658 unit tests, 0 failures.
- Every feature and latency change was built behind a per-commit review gate; the branch additionally passed a whole-branch cross-commit review (the `validatePush` signature threaded from both call sites, B2 and B3a agreeing on the touched prefix, the `markConsumed` result shape consistent with its callers, the off-by-default instrumentation verified inert, no binaries / secrets / debt).
- The latency story is measured, not asserted. Each fix was validated by a fast substrate microbenchmark (committed) and the per-message latency benchmark; one wrong candidate — directory bucketing — was built, measured as a regression, and reverted before shipping. The combined end state is confirmed end-to-end: the FIFO sustained-load test drives 50 mails through the real deploy path to completion in ~47–55s, a workload that previously timed out — so the no-degradation claim rests on a real-path measurement, not the microbenchmark alone.

## Deferred and follow-up

- INTR-214 — the last per-commit O(N) lever (a per-ref isogit index), with the data-corruption gate it requires.
- INTR-210 — a pre-existing reply double-send-on-crash hazard (needs outbound idempotency).
- INTR-215 — a pre-existing timing-fragile child-workflow test.
- INTR-216 — minor branch cleanups.

Resolves INTR-208 and INTR-209.


alexanderguy:
## Reviewer-of-record: whole-branch review

Reviewed `main..HEAD` (32 commits) as a single artifact, plus a commit-message and binary/scope sweep. **Verdict: clean** — one substantive finding fixed in-branch, the rest nonblocking and dispositioned. Each claim below cites the check that backs it.

### Gates
- **Full pipeline green on HEAD:** `make all` → lint + build + 364 tests, 0 fail.
- **Commit-message sweep:** all 32 subjects are imperative, ≤72 chars, prefix-free, unpunctuated, with no tracker / PR / series-position references; bodies are ≤72 with no tracker references. (Three early commit bodies that carried "Phase N / later phase" references were reworded; the trees were verified byte-identical after the rewrite, so only messages changed.)
- **Binary / scope:** `git diff main...HEAD --stat` shows no `Bin` markers on text or source files, no out-of-scope files, and no unexpected dependency or lockfile churn.

### Cross-commit seams verified (what per-commit review cannot see)
- **`KindHandler.validatePush` widening (B3a)** is a backward-compatible optional argument, threaded consistently from both the write-tree and pack-receive call sites; it is `undefined` (validate-all) exactly when there is no parent to diff. Every implementer destructures only what it needs.
- **B2 and B3a agree on the prefix:** `appendBatch` writes with `preservePrefix = runs/<runId>/events/`, which the store converts into exactly the `changedPathPrefixes` set the scoped `validatePush` consumes.
- **`markConsumed` / `MarkConsumedResult` shape (P2)** is consistent with its supervisor caller and the `InboxPrimitives` type.
- **The abandoned bucketing experiment left no residue:** runs are flat `runs/<runId>/`; the deferred per-ref index (INTR-214) is correctly absent.
- **Off-by-default instrumentation** is genuinely inert when its env vars are unset (resolved once at the sidecar boot edge); the three `.bench.ts` files are never matched by the explicit-path test script.
- **Feature fully wired:** routes registered, both admin-UI pages and nav reachable, the migration continues the prior one with no collision, and the per-run disk reclaim is gated to the cold path, targeting a sub-root disjoint from both the warm agent's workspace and the durable conversation store.

### Correctness gates carried per fix
- **B2:** a recovery-equivalence proof — a synchronous run is one commit; a crash mid-segment re-drives exactly once; a crash while suspended resumes from the pre-park flush.
- **B3a:** byte-identical-commit equivalence plus injected-violation-still-rejected, proven at both touched-prefix derivation sites.
- **P2:** exactly-once with no gap — every messageId is covered by either the retained window or the watermark stale-reject, never silently reprocessed; suffix-only prune; monotonic watermark. The replay path deliberately bypasses the stale-reject (documented and tested) so a crashed in-flight message is re-admitted, not lost.
- **Per-run disk-leak reclaim:** the cold-only deletion gate is triple-proven and never wipes a live warm workspace, and the durable conversation store is never touched.

### Findings
- **Fixed in-branch:** the design doc banner still read "not yet implemented" at HEAD; updated to mark the unified single-agent path as built.
- **Nonblocking, accepted:** a redundant read-site default in `markConsumed` (both sides resolve to the same 24h constant).

### Deferred (filed)
INTR-214 (the last per-commit O(N) lever, with the data-corruption gate it needs), INTR-210 (a pre-existing reply double-send-on-crash hazard), INTR-215 (a pre-existing timing-fragile child-workflow test, reproduced on clean `main`), INTR-216 (minor cleanups).

---

### Addendum — commit 33 (`653000c9`, added after this review)

After this whole-branch review, one further commit landed: `653000c9`, in-scope follow-through to B3a. It raises the FIFO sustained-load test to its 50-mail target and rewrites the now-stale throughput comments in the two FIFO test files that still described the receive-side `validatePush` enumeration as unresolved — the only executable change is the `LOAD_MAIL_COUNT` constant; the rest is comment correction.

Reviewed as reviewer-of-record: the diff is one constant plus comment text, no tracker IDs or line-number references, file/symbol refs only; the message is imperative, 65 chars, prefix-free, unpunctuated, body ≤72 with no tracker references. `make all` is green on the new HEAD (`653000c9`) — 3658 unit + 364 integration tests, 0 fail — and the 50-mail run completes in ~47–55s via `bun run test:load`. The **clean verdict holds for all 33 commits.** This also closes the acceptance criterion of INTR-194, now resolved.

alexanderguy:
Rebased onto latest `main` (it had moved 19 commits ahead). Conflicts resolved: our migration renumbered to `0034_sleepy_songbird` (clean regen, no drift); the admin-UI and seed merges keep both main's model surface and this branch's workflow surface; `workflow-substrate-factory` adapted to main's `sources`/`defaultSource` agent-env API; and one redundant commit (a pg-error fix main implemented identically) dropped. The test adaptations were folded into their owning commits, so every commit still builds in isolation. `make all` is green (lint + build + 364 tests, 0 fail), and main's work is intact (model pages and migrations 0031–0033 all present).

alexanderguy:
Added commit `653000c9` — raises the FIFO sustained-load test to its real 50-mail target (`LOAD_MAIL_COUNT` 15→50) and rewrites the now-stale throughput comments in `fifo-mail-load.test.ts` and `fifo-mail.test.ts`.

This is in-scope follow-through to the B3a validatePush scoping (`eef1979a`): that change eliminated the receive-side O(N²) enumeration those comments described as holding the 50-mail target out of reach. Confirmed empirically — `bun run test:load` runs 50 mails to completion in ~47-55s (2 pass, 0 fail), well within the test:load budget. The default `make test` run is unaffected (the load test runs only via `bun run test:load`).

This closes out the acceptance criterion of INTR-194, which is now resolved (the workflow-run validatePush O(N²) it targeted is handled more completely by B3a's scoping).

PR #66, opened 2026-06-24
Wire the approval-flow publish step to the drafted output

alexanderguy:
## Summary

- The seeded `approval-flow` workflow's publish step consumes the drafted release note via an explicit `steps.draft.output` input, rather than inheriting the default-input convention's wiring to the approval gate's signal payload.
- A fixture test asserts publish's input selector resolves to `{ from: "steps.draft.output" }`.

## Verification

- `make all` passes (lint, `tsc -b`, and both test passes; exits 0).
- The fixture still serializes and round-trips through `defineWorkflow` unchanged; publish reads the draft content instead of the null approval signal payload.

Closes INTR-223


alexanderguy:

PR #67, opened 2026-06-26
Compact a terminated run's event files into one combined file

alexanderguy:
## Summary

- A workflow-run repo no longer grows one file per event without bound: when a run terminates, its per-event `runs/<runId>/events/<seq>.json` blobs fold into one combined `runs/<runId>/events.jsonl`, bounding the file count every per-commit git cost scales with. No event is lost — both readers handle either form and the fold is validated byte-for-byte.
- The supervisor learns a run reached a terminal event from a typed handler signal instead of sniffing the committed event path shape, so terminal detection no longer depends on where events live.

## Verification

- `make all` is green (build, lint, 3690 + 364 tests exit 0).
- The fold is gated byte-for-byte: adversarial tests assert a fold that mutates, drops, adds, or reorders an event is rejected, as is a run carrying both forms.
- Form-equivalence tests assert both readers return identical events for a run in either form; a subscribe-replay test confirms a sealed run still replays every event from seq 0.

Closes INTR-214

### DIFF
--- a/packages/hub-sessions/src/index.ts
+++ b/packages/hub-sessions/src/index.ts
@@ -132,6 +132,10 @@ export {
   type WorkflowRunReader,
   type WorkflowRunEvent,
 } from "./workflow-run-reader";
+export {
+  WORKFLOW_RUN_EVENTS_FILE,
+  splitCombinedEventLog,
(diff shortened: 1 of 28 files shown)


alexanderguy:

PR #68, opened 2026-06-26
Make the Makefile the authoritative dev entrypoint and gate the full build

alexanderguy:
## Summary

- The `Makefile` is the single authoritative entrypoint for the
  dev-workflow verbs (`build`, `lint`, `test`, `test-load`, `format`,
  `docs`): each runs its command directly, with tool binaries resolved
  from `node_modules/.bin`. The root `package.json` scripts and the dead
  `packages/db` `db:*` scripts are gone.
- `make build` runs `tsc -b --noEmit --force`, so the gate revalidates the
  entire TypeScript project graph every run instead of trusting incremental
  `tsconfig.tsbuildinfo` state.
- `make all` also runs the admin-ui production bundle via a new
  `build-admin-ui` target (`vite build`), closing the gap where a broken
  Vite bundle merged green — admin-ui's types were already gated, the
  bundler step was not.
- Consumers and docs that named the removed scripts now point at the
  Makefile targets.

## Verification

- `make all` passes (exit 0): lint, forced build, admin-ui bundle, and the
  full suite (336 pass, 28 skip, 0 fail).
- The forced gate catches cross-package breaks: dropping `newlyTerminalRuns`
  from a sidecar `RepoStore` fake makes `make build` exit non-zero, which an
  incremental build passed.
- The admin-ui gate catches bundler-only breaks: a bad import in
  `apps/admin-ui/vite.config.ts` passes `make build` but fails
  `make build-admin-ui` (`Could not resolve ...`, `failed to load config`).
- The pre-commit hook's `make lint` runs clean in its staging checkout.

Closes INTR-230
Closes INTR-231

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -43,6 +43,7 @@ The remaining command gets a clean, running system with seed data. It drops and 
 | Seed (requires running hub)   | `bun bin/seed.ts`              |
 | Full build verification       | `make all`                     |
 | Type check only               | `make build`                   |
+| Bundle the admin UI           | `make build-admin-ui`          |
 | Lint only                     | `make lint`                    |
 | Run tests only                | `make test`                    |
 | Auto-format                   | `make format`                  |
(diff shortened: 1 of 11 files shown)


alexanderguy:

alexanderguy:
Folded in the admin-ui bundle-gate commit (`72b9c982`). Whole-branch `code-review` over all three commits returned clean.

PR #69, opened 2026-06-27
Prevent overlapping inference during pending tool cycles

alexanderguy:
## Summary

- Defer inbound mail behind a tool cycle's in-flight completion events
  so a new inference cannot start mid-batch. A reactor-owned counter of
  outstanding continuation events replaces the history-shape gate, which
  went stale the moment a finished batch's tool results were appended to
  history but before its completion events were consumed.
- Reject a malformed tool sequence (a repeated or orphaned tool result)
  at the inference boundary, failing loud as an internal error that names
  the offending call id instead of an opaque provider HTTP 400.

## Verification

- `make all` passes: lint, type-check, and both test passes are green.
- New tests reproduce the mail-interleave race and assert a single,
  correctly ordered cycle; cover multi-tool batch drain (parallel and
  sequential) and abort-mid-cycle; and exercise the prompt tripwire both
  directly and through the reactor.

Closes INTR-225

### DIFF
--- a/packages/inference/src/reactor.test.ts
+++ b/packages/inference/src/reactor.test.ts
@@ -1796,9 +1796,10 @@ describe("createReactor — abort handling", () => {
 describe("createReactor — dequeue priority", () => {
   test("tool.done is prioritized over message.received when tool calls are pending", async () => {
     // Pre-seed history with an assistant message containing a tool_call
-    // block. Using addToHistory=false on executeTools keeps this as the
-    // last message, so historyHasPendingToolCalls() stays true when the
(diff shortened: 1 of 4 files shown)


alexanderguy:
Self-review returned clean. Each commit passed Greybeard (approach) and
Critique (execution) review under the implement workflow, plus a whole-branch
coherence pass, all clean. Two follow-ups landed afterward and were
re-reviewed clean as a focused delta: an `inference.done`-vs-mail regression
test (failing-first verified) folded into the first commit, and
provider-agnostic wording edits to the tripwire comment and TSDoc. Reviewed at
9104562d.

PR #70, opened 2026-06-28
Make inference provider adapters pluggable through an injected registry

alexanderguy:
## Summary

- Built-in adapters resolve through a synchronous, statically-linked registry; the core inference package's harness, reactor path, and barrel import no concrete adapter.
- Custom third-party adapters are described by an operator-configured manifest of `{provider, specifier, export}` and loaded by dynamic `import()`. The harness resolves through `Dependencies.adapters`; `createAgent` fills `env.deps` with a sync built-in default when it is omitted.
- The sidecar composes its registry once at the boot edge, uses it on the in-process single-agent path, and serializes the manifest into the forked workflow child, which re-validates the shape and rebuilds an equivalent registry by re-importing the specifiers — so a custom adapter resolves in every process an agent runs in. Specifiers come only from operator config; deploy and tenant config carry only a `provider` key, so a provider string cannot introduce a module to import.

## Verification

- Lint, type-check (`tsc -b --noEmit --force`), and the admin-ui production bundle pass.
- All adapter tests pass, including a real forked-child integration test (`tests/workflow-deploy/cross-process-custom-adapter.test.ts`) that resolves a manifest-supplied custom adapter inside the spawned child and a negative control proving an unmanifested provider is rejected at the source gate.
- One pre-existing, load-sensitive timeout test in `hub-sessions` claim-check (untouched by this branch; fails on this machine at the pre-merge baseline under sustained load; passes with a longer timeout) is the only failing test in a local `make all` run. It is unrelated to inference adapters.

Closes INTR-233


alexanderguy:
Self-review (`code-review` skill) returned clean. Whole-branch review (reviewer-of-record commit/stat/message audits + a `critique` deep read of `main...HEAD`) surfaced one low-severity finding — a stale comment referencing the deleted `registerProvider` API — which is fixed and autosquashed onto the commit that orphaned it. No outstanding findings or waivers.

PR #71, opened 2026-06-28
Decouple hub routing and availability from sidecar session restore

alexanderguy:
## Summary

- On socket-open the sidecar registers for routing with an empty address
  list, so a provision arriving while it restores sessions routes
  immediately instead of failing with "No sidecar available"; the hub
  keeps that provision's routing and connector state across the
  restore-completing reconnect.
- The registry tarball fetch carries a deadline spanning the request and
  the streamed body read, so a stalled registry fails the fetch instead
  of hanging session restore.
- Session restore runs agents with bounded parallelism, so one slow or
  failing agent no longer blocks restoration of the others.

## Verification

- `make all` passes: lint, type-check, the admin-ui production bundle,
  and both test suites exit 0 (3732 + 339 tests, 0 failures).
- Tests cover the new behavior: a provision routed during restore stays
  routable and keeps its connector state across the reconnect; a stalled
  registry fetch aborts at its deadline; bounded-parallel restore
  isolates a wedged agent while healthy agents restore.

Closes INTR-235

### DIFF
--- a/packages/hub-agent/src/session-manager.test.ts
+++ b/packages/hub-agent/src/session-manager.test.ts
@@ -469,3 +469,168 @@ describe("SessionManager.onAgentEvent per-agent fan-out", () => {
     expect(bobEvents).toEqual([bobTick, bobTick]);
   });
 });
+
+async function seedAgentsOnDisk(
+  dataDir: string,
+  addresses: string[],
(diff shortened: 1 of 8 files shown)


alexanderguy:

PR #72, opened 2026-06-29
Materialize tool packages at a stable per-deploy-id path

alexanderguy:
## Summary

- Tool-package applies materialize each deploy into a stable
  `packages/<deploy-id>/` directory that is never renamed; a single
  `active-deploy-id` file write is the commit point.
- Loaded modules keep a valid on-disk URL for the life of the deploy, so a
  tool package can resolve its own files at call time (`import.meta.url`,
  `require.resolve`, `await import("./sibling.js")`).
- A prelude sweep retains only the current and previous deploys, bounding
  disk while keeping the prior tree resolvable for a session still draining
  against it.

## Verification

- `make all` passes: lint, type-check, admin-ui bundle, and both test
  passes (exit 0).
- `loader.test.ts` drives the real native importer end-to-end — a retained
  previous deploy resolves a call-time sibling import, and a reaped deploy
  does not.

Closes INTR-176

### DIFF
--- a/apps/sidecar/src/tool-materialization.test.ts
+++ b/apps/sidecar/src/tool-materialization.test.ts
@@ -292,7 +292,7 @@ describe("materializeToolPackages — manifest.invalid gate", () => {
 
   test("boot reconciliation reads the dirty marker as previousDeployId", async () => {
     // Seed an instance where the prior apply could not durably record
-    // the post-swap deploy id and a dirty marker carries the truth.
+    // the committed deploy id and a dirty marker carries the truth.
     // The next apply (any apply — here, a manifest-invalid one) must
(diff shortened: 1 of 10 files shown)


alexanderguy:

- `apps/sidecar/src/tool-materialization.ts:395-425`: the degraded-persist path that emits `apply.previous-rotation.failed` carrying the new live deploy id in `previousDeployId` has no end-to-end integration test; its degradation ladder and boot reconciliation are unit-tested and the frame field is a pass-through. Authorized by Greybeard on grounds of churn-avoidance — the emission logic is pre-existing and unchanged, and an end-to-end test would reverse a documented anti-fs-injection decision in the test file and require successful-apply fixtures that do not exist.

PR #73, opened 2026-06-29
Drain in-flight repo operations before deleting an agent directory

alexanderguy:
## Summary

- Routes the sidecar agent repo's git operations — mail-audit commits,
  state-pack and deploy-ref reads, and deploy/asset-pack applies — through
  one per-agent serialization chain.
- Drains that chain in `deleteAgentDir` before removing the directory, so a
  state-pack read racing an `agent.undeploy` finishes before the `rm` instead
  of failing against a vanished path.
- Keeps the hub's structured-rejection catch as a defence-in-depth boundary;
  the directory-teardown case no longer reaches it.

## Verification

- `make all` passes (exits 0): lint, type-check, admin-ui bundle, and tests.
- A session-manager unit test asserts `deleteAgentDir` removes the directory
  only after an in-flight read completes, and that a rejecting operation
  propagates to its caller without poisoning the chain.
- The deploy-flow integration test races a sync-request state read against
  `endSession`: the sidecar directory is removed, agent state still pushes to
  the hub, and no state-push read failure is logged.

Closes INTR-150

### DIFF
--- a/packages/hub-agent/src/session-manager.test.ts
+++ b/packages/hub-agent/src/session-manager.test.ts
@@ -14,7 +14,7 @@ import type {
 } from "@intx/types/runtime";
 
 import { createAgentKeyStore } from "./agent-key-store";
-import { createAgentRepoStore } from "./agent-repo-store";
+import { createAgentRepoStore, type AgentRepoStore } from "./agent-repo-store";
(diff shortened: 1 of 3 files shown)


alexanderguy:

PR #74, opened 2026-06-29
Garbage-collect agent repos on the hub and sidecar write paths

alexanderguy:
## Summary

- A collector (`runGC` in `@intx/storage-isogit`) repacks everything reachable from an agent repo's refs into a single pack and drops the superseded packs and loose objects. The keep set unions reachability over both `refs/heads/main` and `refs/heads/deploy`, and a retention policy selects full ancestry (`keep-history`) or current tips only (`tip-only`).
- The hub and sidecar reclaim on the write path through one shared evaluator, gated by pack-count and loose-object thresholds, and emit a disk-pressure warning when a repo crosses a byte budget. The hub triggers after state-pack receives and deploy commits under its substrate lock; the sidecar triggers from the reactor's context commits.
- A per-directory lock makes the storage layer the single owner of object-store concurrency for an agent repo, so the collector never races the reactor's commits, the mail-audit commits, or a deploy apply.
- Retention is per environment: the hub is fixed to `keep-history` (it is the long-term archive and its history replay depends on the ancestry), while the sidecar defaults to `tip-only` for small repos. The durable conversation lives in the working-tree files at the tip, so the sidecar's history-read paths degrade to the surviving slice rather than failing under a pruned ancestry.
- Rejected packs are still unpublished at the moment of rejection, so a flood of bad packs cannot accumulate before the next threshold-driven collection.

## Verification

- `make all` passes: lint, `tsc -b` across the graph, the admin-ui production bundle, and the test suites (3735 unit tests and 339 integration tests, 0 failures).
- The GC unit tests pin both-ref reachability preservation, `tip-only` vs `keep-history` retention, dangling-parent tolerance on the deploy ref, commit-versus-collector serialization under the lock, and history-read degradation after a `tip-only` collection.
- Every commit on the branch type-checks independently (`make build`).

Closes INTR-151


alexanderguy:

PR #75, opened 2026-06-29
Migrate the interchange crypto surface to the Web Crypto API

alexanderguy:
## Summary

- The crypto package (`@intx/crypto`, renamed from `@intx/crypto-node`) runs on
  `globalThis.crypto.subtle` for Ed25519 keygen/sign/verify, key import via the
  PKCS#8/SPKI DER prefixes, and SHA-512 — with no `node:crypto` or `Buffer`
  dependency. Its sign/verify/import surface is async and is threaded through
  every caller across the workspace. The provider is exported as
  `Ed25519Crypto` / `createEd25519Crypto`, named for its algorithm rather than
  its runtime.
- OpenPGP detached-signature encoding lives in one place: the crypto package
  exposes a signer-function primitive, and the MIME package delegates to it
  instead of carrying a duplicate copy.
- No non-crypto package imports `node:crypto`: identifier generation uses the
  Web Crypto global, SHA-256 digests use `subtle.digest`, and the workflow-host
  IPC channels authenticate with Web Crypto Ed25519 and HMAC-SHA256 under a
  constant-time tag comparison.
- The sidecar validates its signing keypair on load: it derives the public key
  from the seed and fails loudly when the on-disk public key disagrees, so a
  corrupted or swapped key file is caught at startup rather than as a confusing
  hub-side verification failure.
- `@typescript-eslint/no-floating-promises` and `no-misused-promises` are
  enabled repo-wide, mechanically guarding the async sign/verify surface against
  a future dropped `await`.

## Verification

- `make all` passes (0 failures) with the new lint rules active, on HEAD rebased
  onto `origin/main`.
- Wire format is unchanged: committed known-answer fixtures pin byte-identical
  Ed25519 signatures, armored PGP packets, and SSHSIG blobs captured from the
  pre-port Node implementation; the new code reproduces them and verifies the
  pre-port signatures.
- Coverage includes golden SHA-256 byte-stability locks (git-token and blob
  hashes), control-channel bootstrap negative branches and sender-serialization
  wedge resistance, the constant-time HMAC compare, and fail-closed sidecar
  keypair validation.
- `rg 'from "node:crypto"'` / `rg 'from "crypto"'` / `rg 'NodeCrypto'` over
  `packages/ apps/ examples/` return no results.

Closes INTR-135
Closes INTR-136
Closes INTR-137
Closes INTR-238
Closes INTR-251


alexanderguy:

alexanderguy:
Follow-up: this branch now also folds in INTR-238 (rename the provider to `Ed25519Crypto`/`createEd25519Crypto`) and INTR-251 (validate the sidecar signing keypair against its seed on load, failing closed on a mismatched on-disk public key). The four added commits each passed `make all`, Greybeard review, and per-commit Critique; a fresh whole-branch `code-review` over `origin/main...HEAD` returned clean.

alexanderguy:
Coverage + hardening follow-up (commits `5cb4ba7f`, `23d5417b`, `f5861dd3`):

- Added the test gaps a coverage review surfaced: golden SHA-256 byte-stability locks on the git-token hash (on the production-bound helper) and the blob-substrate hash; control-channel bootstrap negative branches (non-`ready` first frame, malformed pubkey hex, key-pinning); and a control-sender rejecting-write wedge test.
- Enabled `@typescript-eslint/no-floating-promises` and `no-misused-promises` repo-wide and cleared all 110 violations (96 fire-and-forget `void`, 14 awaited test/bench teardowns — no production missed-awaits; the crypto/verify/sign paths had zero). This mechanically guards the async sign/verify surface against a future dropped `await`.

Each new commit passed `make all` (with the rules active), Greybeard review, and per-commit Critique.

PR #76, opened 2026-06-30
Replace drizzle-introspection test mocks with a real-database harness

alexanderguy:
## Summary

- Adds a private `@intx/test-harness` workspace package: a migrate-once-per-file
  Postgres harness (`createTestDb`) with truncate-between-tests isolation, plus
  fixture seed helpers, gated on database env so a checkout without it still runs
  `make all`.
- The asset, tenant-hierarchy, catalog, and model-source resolver suites and the
  hub-api inherited-asset and tarball route tests now run their real drizzle
  queries against a migrated throwaway schema; the harness replaces a stub that
  introspected drizzle's internal query AST.
- New `credential-resolution` tests cover the `isNull`, scope-containment, and
  ambiguity paths the stub could not express.

## Verification

- `make all` exits 0: lint, type-check (`tsc -b`), the admin-ui production
  bundle, and both test passes.
- The new DB suites run against a real Postgres schema when database env is
  present and skip cleanly when it is absent.
- All introspection mocks are removed, so no surviving mock needed a pointer
  comment to the harness.

Closes INTR-180


alexanderguy:

PR #77, opened 2026-06-30
Keep the database and orchestrator out of the workflow-child boot graph

alexanderguy:
## Summary

- The spawned workflow-child's boot module graph no longer evaluates the
  control-plane database layer (`@intx/db`/drizzle) or the sidecar
  orchestrator (`@intx/hub-agent`). Both packages are reached only through
  narrow db-free subpath entries — `@intx/hub-sessions/substrate` and
  `@intx/hub-agent/paths` — and the package barrels are unchanged for
  hub-side consumers.
- `@intx/hub-sessions/substrate` exposes the package's db-free substrate
  layer (workflow-run claim-check, event-log, repo-store protocol, and the
  workflow-definition schema); `@intx/hub-agent/paths` exposes the
  deploy-tree reader and address helper without the orchestrator's module
  graph.
- A hub-api test imports the crypto provider under its current package
  name `@intx/crypto`.

## Verification

- `make all` passes: lint, `tsc -b --noEmit --force`, the admin-ui
  production bundle, and tests (3696 + 339 tests, 0 failures).
- A module-load probe importing both workflow-child roots confirms
  `@intx/db`, `drizzle-orm`, and `@intx/hub-agent` are absent from the
  child's module graph (each re-imports cold; the `@intx/workflow` control
  re-imports cached).

Closes INTR-254

### DIFF
--- a/apps/sidecar/src/conversation-state.ts
+++ b/apps/sidecar/src/conversation-state.ts
@@ -130,8 +130,12 @@ import { type } from "arktype";
 import { getLogger } from "@intx/log";
 import { createConnectorRouter } from "@intx/harness";
 import { createIsogitStore } from "@intx/storage-isogit";
-import type { Principal, RepoId, RepoStore } from "@intx/hub-sessions";
-import { WORKFLOW_RUN_AGENT_STATE_PREFIX } from "@intx/hub-sessions";
+import type {
+  Principal,
+  RepoId,
+  RepoStore,
+} from "@intx/hub-sessions/substrate";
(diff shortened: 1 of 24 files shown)


alexanderguy:

- `packages/hub-sessions/src/substrate.ts` and `packages/hub-sessions/src/index.ts` independently enumerate the overlapping substrate-layer symbol set, rather than the barrel deriving it from the `/substrate` entry. Authorized by Greybeard on the grounds that the two are distinct contracts — a curated, hand-audited db-free boot boundary versus the package's full public surface — and a boundary allowlist should remain an explicit enumeration rather than be auto-derived from the surface it constrains.

PR #78, opened 2026-06-30
Replace Node Buffer with Web-standard codecs in non-crypto code

alexanderguy:
## Summary

- Non-crypto packages, tests, and `bin` scripts contain no Node-only `Buffer`; byte handling runs on `Uint8Array` and the shared `@intx/types` codecs (`base64Encode`/`base64Decode`, `hexEncode`/`hexDecode`, the new `base64urlEncode`/`base64urlDecode`, and an N-ary `concatBytes`), with inline `TextEncoder`/`TextDecoder` for the utf8 leg.
- The `db` bytea columns, the tarball cache/loader, and the package-json extractor carry `Uint8Array` end to end; the pagination cursor and git PAT secret keep their exact base64url wire format through the new codec.
- Modules that stay Node-bound for other reasons (`tar`, `node:stream`, `child_process`) are documented as such, so the absence of `Buffer` is not mistaken for portability.

## Verification

- `make all` passes (lint, `tsc -b --force`, admin-ui bundle, tests): unit suite 3711 pass / 0 fail, integration suite 444 pass / 28 skip / 0 fail.
- base64url encode/decode is byte-compatible with Node's `"base64url"` across lengths 0–300 and round-trips (fixed-vector test); malformed input throws at the codec and is caught at each decode boundary.
- The `db` bytea round-trip is exercised against a real database across empty, high-byte, and embedded-NUL inputs.
- No `Buffer` usage remains outside `packages/crypto` (tree-wide grep; only doc comments mention it).

Closes INTR-253

### DIFF
--- a/apps/sidecar/src/agent-key-registration-lifecycle.test.ts
+++ b/apps/sidecar/src/agent-key-registration-lifecycle.test.ts
@@ -16,6 +16,7 @@ import os from "node:os";
 import path from "node:path";
 
 import { createEd25519Crypto, generateKeyPair } from "@intx/crypto";
+import { hexEncode } from "@intx/types";
 import { createInMemoryTransport } from "@intx/mail-memory";
 import type { RepoId, RepoStore } from "@intx/hub-sessions";
 import {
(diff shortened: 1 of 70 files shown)


alexanderguy:

PR #80, opened 2026-07-01
Encode package-qualified tool names for provider wire charsets

alexanderguy:
## Summary

- Add an invertible codec that rewrites tool names into a provider's
  function-name charset and inverts them on the way back, so
  package-qualified ids like `@intx/tools-posix/sidecar-bundle:run_shell`
  reach DeepSeek, Kimi/Moonshot, Anthropic, and Gemini as valid names.
- Wire the codec into the OpenAI, Anthropic, and Google GenAI adapters at
  every tool-name emission site and decode at every parse site, so
  tool-call dispatch still keys on the original prefixed name.
- Leave already-valid names untouched; only names with out-of-charset
  characters or a leading non-letter are rewritten, and an over-long wire
  name throws rather than reaching the provider as an opaque 400.

## Verification

- `make all` passes (lint, `tsc -b`, admin-ui bundle, tests); exits 0.
- Each adapter carries a round-trip test proving a package-qualified name
  survives buildRequest → parseResponse back to the prefixed form; the
  codec has unit coverage for passthrough, rewrite, marker collision,
  leading digit, the over-limit throw, and total decode.

Closes INTR-217

### DIFF
--- a/packages/inference-testing/src/wire/openai.test.ts
+++ b/packages/inference-testing/src/wire/openai.test.ts
@@ -1,4 +1,5 @@
 import { describe, test, expect } from "bun:test";
+import { type } from "arktype";
 
 import { parseSSE } from "@intx/inference";
 import { createOpenAIAdapter } from "@intx/inference/providers";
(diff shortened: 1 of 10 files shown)


alexanderguy:

PR #81, opened 2026-07-05
Land the workflow-run substrate latency fix and remove the trivial agent path

alexanderguy:
## Summary

Lands the workflow-run substrate latency fix and removes the trivial
in-process agent path, then builds live source rotation on top of the
resulting supervised single-step deployment and cleans up the wire
surface the removals left behind.

- **Flattens per-message substrate cost.** Delta-scoped claim-check
  validation becomes the production default, repo-store commit trees
  assemble without the on-disk git index, a per-repo object cache backs
  the git calls, and the substrate mirror slices new turns from the
  reactor's in-memory array instead of reloading the conversation from
  disk each message. Per-message growth flattens and the dispatch loop
  no longer wedges as history grows; exactly-once is preserved and the
  index-free tree assembly is proven byte-identical to canonical git.

- **Removes the trivial in-process agent path.** Single-agent instances
  now deploy as single-step workflows at the head. The in-process
  session runtime, the warm-harness launch entry point, the session-start
  wire path, and the dead agent-metadata persistence layer are retired,
  and the orchestrator's trivial deploy branch is deleted.

- **Rotates inference sources on a running deployment.** A supervised
  single-step child swaps its warm agent's sources in place on a
  `sources-updated` control frame, with a mutable sources reference
  covering the cold pre-first-build window. The rotation routes
  hub → sidecar → supervisor → child, and the sidecar answers the inbound
  `sources.update` with `session.ack` / `session.error`.

- **Hardens the supervised deployment lifecycle.** A spawn that throws
  after the child process starts but before the ready handshake routes
  through the single teardown owner, so a failed startup neither orphans
  the child process nor leaks its mail subscription. A source rotation is
  persisted to the deployment record before it is committed in memory, so
  a failed persist leaves no partial rotation, and the rotation survives
  both a recycle respawn and a full sidecar restart.

- **Hardens deploy against concurrent same-address deploys.** An
  in-flight reservation rejects a second deploy for an address still
  being provisioned, and a failed single-step spawn reverses the
  recorded hub key so a retry starts clean.

- **Answers malformed inbound control frames.** A malformed request or
  pack frame is answered through its correlation key
  (`session.error` / `agent.error` / `repo.pack.reject`) so the hub's
  request no longer hangs to its timeout instead of being silently
  dropped.

- **Retires the orphaned `session.abort` / `grants.update` wire
  surface.** Retiring the in-process session runtime left the hub able to
  send these two frames with no live sidecar receiver. Their senders,
  frame types, and test mocks are removed, and the docs that described the
  retired grant/abort mechanism are corrected.

- **Serializes the substrate-mirror entry points.** The awaited
  `onRunBoundary` mirror and the fire-and-forget `onStateChanged` mirror
  shared the mirrored-count state with no serialization; both, and the
  restore that can re-enter the mirror, now run on one per-instance tail
  so overlapping runs cannot collide on a boundary and drop a turn.

## Verification

- `make all` is green on HEAD: lint (prettier + eslint, including
  `gen-api-docs --check`), build, admin-ui build, and the full test
  target — 3814 package/unit tests and 344 integration tests pass with
  zero failures.
- The branch is reviewed as a whole against `main` over its full range,
  and each commit was reviewed as it landed. The reviewer-of-record
  audits are clean: commit-message style across every commit (no
  prefixes, filenames, tracker references, cross-commit references, or
  over-length lines), a `--stat` scan for binary or out-of-scope files,
  and the `main..HEAD` scope.
- The whole-branch review found no correctness defects. The findings it
  surfaced were comment and documentation staleness left by the
  removals; the branch-caused ones are corrected here (the trivial-path
  comment residue and the harness-design restore-model doc), and the
  remaining lower-priority residue is tracked in INTR-264.
- The supervised-lifecycle and rotation-durability fixes are covered by
  regression tests verified to fail without the fix, including the
  spawn-teardown subscription release, the first-build-vs-rotation race,
  and the persist-fails-leaves-no-partial-effect atomicity guard.
- The two substrate-mirror concurrency defects are covered by
  barrier-driven regression tests verified to fail without the fix.

## Follow-ups

- INTR-261 — the dead `session.abort` / `grants.update` wire surface is
  removed on this branch. What remains under the issue is *deciding the
  supervised replacement* for each retired semantic (instant live grant
  re-point; a supervisor cancel/drain for abort), which is a design
  decision, not cleanup.
- INTR-257 — reject base-type / delete-shape mismatches in the delta
  tree write (guard implemented on this branch).
- INTR-264 — reconcile the remaining stale architecture docs and
  comments left by the trivial/in-process removal (the
  unified-execution-host design doc, the harness-design package-structure
  section, and a dead supervisor re-export). Documentation only; no
  correctness change.

Closes INTR-256
Closes INTR-262


alexanderguy:
## Remediation update: fixes from a full-branch review

This pushes 15 commits (`b00e9f7f..HEAD`, +748/−520 across 31 files) addressing findings from a full-branch code review. Each behavioral change ships with a mutation-verified regression test; every finding is dispositioned (fixed or verified non-issue).

### Error-path fixes (behavioral)
- **Reconnect deploy-ref read** no longer silently drops a deployment's routes: a failed read logs and closes the socket to force a clean reconnect (restoring behavior that had regressed).
- **Recycle respawn** reaps the freshly-spawned child instead of leaking its process + IPC channels when the pre-handshake credentials read throws.
- **Spawn failure** preserves the original cause when teardown also throws.
- **`shutdownInternal` is structurally total**: the child kill and the transition to `stopped` live in a `finally`, so the supervisor always kills the child and always reaches `stopped` regardless of which teardown step throws; the terminal broadcaster's `dispose` is made total at its owner.
- **`sources.update` boundary** rejects an empty `sources` list (matching the deploy frame's per-step arrays and the live producer, which never emits empty).

### Dead-code removal
- Removes the fully-dead `deploy.apply.error` frame + emitter subsystem (−238 lines net). `DeployApplyErrorCategory` (the live tool-packaging error taxonomy) is retained.

### Documentation / comment accuracy
- Corrects ~a dozen comments and docs to match shipped code: the delta claim-check mechanism (`writeTreeDelta` vs the old wholesale-rewrite), the in-process-harness removal references across three packages, the migration/schema `NOT NULL` contradiction, the source-rotation rollback rationale (attributed to the sidecar frame queue, not hub ack-pacing that does not exist), and stale symbol fates in the execution-host design doc.

### Verification
- `make all` green on HEAD: **3840 main + 480 DB tests, 0 fail**.
- Whole-branch review clean (cross-commit findings fixed); commit-message + binary + scope audits clean.
- No follow-up issues deferred; no open TODOs.

PR #82, opened 2026-07-09
Add native loop and effect primitives to the workflow engine

alexanderguy:
## Summary

- Add an `action` primitive: deterministic, capability-checked effect nodes
  backed by an effect ledger, so git and build work runs crash-safe and
  exactly-once without an agent in front of it.
- Add a bounded `loop` primitive: a child run per iteration with pure
  while/carry registry functions, an iteration cap, and onExhausted routing,
  resumable to exactly-once across a mid-loop crash.
- Reject dependency cycles at definition time, and reject a duplicate child
  spawn in the state machine.
- Run the interchange-demo per-level pipeline — plan, implementer map, commit,
  gate, amendment loop — end-to-end on the engine with real agents through the
  production step-invoker.

## Verification

- `make all` passes: build, lint, and both test passes exit 0; the
  workflow-deploy suite runs the real-agent pipeline test.
- The crash-safe exactly-once guarantees carry standing regression tests
  (action, effect-context, loop-resume, loop-iteration, and the dispatch
  demo), not just a spike.
- The real-agent pipeline test drives the composition end-to-end through the
  production step-invoker: a mid-loop crash resumes and re-drives the final
  amendment iteration, and the shared ledger holds its rebuild effect to one
  execution. A defeated-ledger probe (whose lookup never dedups) reruns the
  same crash and asserts the effect count rises, proving the exactly-once
  assertion is not a durable-log replay.
- The real-agent test fans the pipeline over two tasks and asserts each map's
  inner agent receives and returns its own task, so the fan-out threads
  per-task input rather than serving one task's output to both.

Closes INTR-270


alexanderguy:

PR #83, opened 2026-07-10
Recover deployed workflows across sidecar reconnection and crash-restart

alexanderguy:
## Summary

- Fix three latent reconnection defects: a permanent deploy-window wedge where a failed reconnect challenge forgot the sidecar's on-disk key, silent dropping of disconnect-queue mail, and workflow-run pack data loss when a transfer is interrupted mid-flight.
- Recover an interrupted workflow-run pack after reconnect (drain and re-drive the push once the address is routable again, and register the deployment address before the supervisor replays), and resume a workflow parked at an await-signal gate after a sidecar crash (a runtime resume carve-out plus a restore-path guard that suppresses the redundant re-trigger); the runtime fails loud rather than binding a payload to the wrong gate when two concurrent same-name gates are ambiguous.
- Pin reconnection and crash-restart survival with end-to-end integration tests that drop or crash a real sidecar subprocess and assert exactly-once recovery, document pinned-forever workflow-definition versioning, and add observational deploy-latency and boot-restore benchmarks.

## Verification

- `make all` passes: `bun test` exits 0 across the suite (3919 + 361 tests, 0 failures).
- The six reconnection and restore integration tests each stand up a real sidecar subprocess, sever the hub link or kill the process, and assert the deployment recovers and the run completes with effects applied exactly once.

Closes INTR-277


alexanderguy:
Self-review (`code-review` skill): the whole-branch review returned no blockers. The one warning — a silent wrong-payload on resume for the ambiguous concurrent same-name await-signal topology — is fixed with a fail-loud guard, and the review nits are fixed. No outstanding exceptions. Follow-ups filed: INTR-280 (mid-step at-least-once) and INTR-281 (concurrent same-name gate validator).

PR #84, opened 2026-07-13
Make agent-step side effects at-most-once across a mid-step crash

alexanderguy:
## Summary

- An agent step flushes its `StepStarted` durably before the agent runs, so a
  crash mid-invocation leaves a durable marker; on recovery the crashed step
  settles as a terminal failure instead of the agent being silently re-invoked.
  Side effects are at-most-once across a mid-step crash rather than at-least-once.
  Exactly-once is unattainable for a non-deterministic agent, so the failure is
  loud and operator-visible.
- Crash recovery reduces canonical state from the durable log, so a supervisor
  re-fire that carries no seed adopts the surviving log — settling a crashed step,
  or short-circuiting an already-terminal run — instead of stalling.
- Each run is driven by a single driver on the child: a `trigger.fire` for a run
  self-discovery is already resuming is declined rather than spawning a second
  concurrent driver.

## Verification

`make all` exits 0 (lint, type-check, admin-ui bundle, and both test passes). New
runtime tests cover crash-mid-agent-step recovery (agent invoked at most once, run
settles `RunFailed` with a `crash-mid-invocation` reason), the no-seed adopted-log
re-fire (settle-and-fail and the terminal short-circuit across completed/failed/
cancelled), and the seeded adopt-by-skip case; a workflow-host test covers the
single-driver claim under a re-fired trigger.

Closes INTR-280

### DIFF
--- a/packages/workflow-host/src/child/run-child.test.ts
+++ b/packages/workflow-host/src/child/run-child.test.ts
@@ -890,6 +890,221 @@ describe("runWorkflowChild", () => {
     expect(result.resumedRunIds).not.toContain("run-done");
   });
 
+  test("a re-fired trigger for a self-discovered run does not spawn a second driver", async () => {
+    // Commit 1 leaves a crashed agent step as a discoverable non-terminal
+    // log; the supervisor both resumes it (self-discovery) AND re-fires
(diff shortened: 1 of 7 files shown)


alexanderguy:

PR #85, opened 2026-07-13
Flush an action's StepStarted durably before invoking it

alexanderguy:
## Summary

- `runAction` flushes the action's `StepStarted` durably before
  `env.invokeAction`, mirroring the agent-step barrier in `runStep`, so a
  lone action that crashes mid-invocation leaves a durable residual.
- The crash-recovery pass settles that residual as a terminal
  `StepFailed` (`crash-mid-invocation`) instead of silently re-invoking
  the handler, giving actions the same at-most-once crash semantics as
  agent steps.

## Verification

- `make all` passes: lint, `tsc -b --force`, the admin-ui bundle, and
  both test passes exit 0.
- `resume-mid-action-step.test.ts` captures the durable log at the
  instant of invocation and resumes from it, asserting the action is
  invoked at most once and the run settles `RunFailed`; the test fails
  when the barrier is reverted.

Closes INTR-284

### DIFF
--- a/packages/workflow/src/runtime/dag.ts
+++ b/packages/workflow/src/runtime/dag.ts
@@ -93,22 +93,17 @@ export function isResumableReceivedAwaitSignalStep(
 /**
  * An `in-flight` step whose primitive is an invocation boundary -- an
  * agent `step` or a deterministic `action` -- is a crash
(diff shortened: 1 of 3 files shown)


alexanderguy:

PR #87, opened 2026-07-14
Authenticate the sidecar WebSocket connection

alexanderguy:
## Summary

- The sidecar WebSocket handshake requires a per-sidecar bearer token,
  validated against a SHA-256 hash stored on the `sidecar` table; the hub
  keys connection state off the token-verified identity, not the claimed
  frame value.
- Sidecar registration is WebSocket-only; the hub exposes no
  `/api/sidecars` REST surface, so a routable sidecar identity cannot be
  created without a provisioned token.
- `bin/dev.ts` provisions the dev sidecar's identity before it connects,
  and the trust-model docs state the enforced model: a per-sidecar token
  on the handshake plus the Ed25519 reconnect challenge, with mutual TLS
  and network isolation left to deployment rather than enforced in code.

## Verification

- `make lint` and `make build` pass (`tsc -b` clean); the unit and
  integration suites pass, including deploy and reconnect integration
  tests that exercise the token-authenticated handshake end-to-end with a
  token on the wire.
- A real-DB regression test (`tests/db/sidecar-token-authenticator.test.ts`
  on the `createTestDb` harness) drives the shipped
  `createSidecarTokenAuthenticator` against a migrated Postgres schema,
  covering the `bytea` token-hash write-then-read round-trip: a valid token
  resolves to the seeded identity, wrong and empty tokens resolve to
  `null`, a spoofed claimed id with a valid token resolves to the token's
  verified id, and among several rows each token selects its own row by
  hash. It runs in CI and skips when no database is configured.
- `resolveProvisionConfig` boundary validation is unit-tested directly.

Closes INTR-259


alexanderguy:
Self-review (`code-review` skill) is clean. The security property holds against source: routing keys off the token-verified identity (never the claimed frame value); auth is required and fail-closed on both the register and reconnect handshakes; the allow-all test helper is test-only and unreachable from production; and the deploy and reconnect integration tests drive the real authenticator with a token on the wire. Gate green (`make lint`/`make build` exit 0; unit and integration suites pass).

alexanderguy:
The DB write path is verified against a live Postgres (a throwaway scratch database; the primary dev DB is untouched):

- The full migration chain (0001 through 0036) applies cleanly; `0036` adds `token_hash_sha256 bytea NOT NULL` plus the unique constraint, and the not-null column applies to the empty `sidecar` table.
- `bin/provision-sidecar.ts` writes the sidecar row with the stored hash equal to `sha256(token)`.
- The shipped `createSidecarTokenAuthenticator` against the live row resolves a valid token to its verified identity, resolves a wrong or empty token to `null`, and resolves a spoofed claimed `sidecarId` carrying a valid token to the token's verified id (never the claim).
- Provisioning surfaces the DB error and exits non-zero when the column is absent, rather than swallowing it.

PR #88, opened 2026-07-14
Emit valid action sets on the director abort and halt paths

alexanderguy:
## Summary

- `DefaultDirector.afterInferenceDone` emits `[checkpoint, done]` on an
  abort decision and `[checkpoint, reply]` on a halt decision — action
  sets the reactor accepts, so an abort terminates the session and a halt
  pauses it and replies, instead of crashing the reactor with a fatal
  "Invalid action set".
- The abort path does not surface its `reason`: a reply invites
  continuation, which a terminating action cannot honor. The `reason`
  remains a policy input.
- The director header comment scopes its event map to inbound events and
  points the `afterInferenceDone` continue/abort/halt decisions at the
  `AfterInferenceDecision` contract.
- Tests lock the `reply`+`done` and `wait`+`reply` validator invariants
  for the director's checkpoint-prefixed action shape, and drive the real
  director through the reactor to prove abort terminates cleanly and halt
  replies while staying alive.

## Verification

- The full build passes: `make all` exits clean (unit 3956 pass / 0 fail,
  integration 361 pass / 0 fail).
- The abort path reaches `reactor.done` and the halt path emits
  `connector.reply` while staying alive, with no fatal "Invalid action
  set" error in either case.

Closes INTR-201

### DIFF
--- a/packages/inference/src/default-director.test.ts
+++ b/packages/inference/src/default-director.test.ts
@@ -20,6 +20,7 @@ import type {
   DefaultDirectorPolicy,
 } from "./default-director";
 import { createCapabilities } from "./director";
+import { validateActions } from "./actions";
 
 // ---------------------------------------------------------------------------
 // Fixtures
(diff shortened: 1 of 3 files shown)


alexanderguy:

PR #89, opened 2026-07-15
Distribute the @intx packages as compiled ESM

alexanderguy:
## Summary

- Published `@intx/*` packages ship compiled ESM (`dist/`, `.js` + `.d.ts`)
  instead of TypeScript source, so a consumer on any ESM runtime installs
  and runs them without a TypeScript toolchain. A custom `intx-src` exports
  condition keeps the in-repo dev loop on source while consumers get `dist`.
- Every in-repo launcher that resolves `@intx/*` carries the `intx-src`
  condition — the `bin/` wrappers, the sidecar's spawned `workflow-child`,
  and the example `start` scripts — and a `check-launchers` lint guard fails
  the build if a launcher drops it or a `bin/` entry point's wrapper shadows
  its `.ts`.
- `bin/publish` is a guarded, dry-run-by-default publish path: it requires a
  clean tree at the release tag, version-synced packages, `workspace:`/
  `catalog:` internal deps, and a fresh `bun.lock`, then packs every package
  and asserts each package root and every exports subpath loads under Node,
  Bun, and Deno resolving to `dist` (not the inert `intx-src` source).
  `bin/release` refreshes the workspace versions `bun.lock` records so
  `bun pm pack` ships the release version.
- `vscode-jsonrpc@9` lets `@intx/tools-lsp` load under Node and Deno, so the
  whole distribution runs on all three runtimes.

## Verification

- `make all` passes (build, lint, admin UI, 4060 + 361 tests).
- `make verify-tool-load` loads the tool closure from a packed 9-package set
  under plain `bun`, resolving `@intx/*` to compiled `dist`.
- `check-launchers` fails lint when a launcher resolves `@intx/*` without the
  condition, verified by stripping a flag from a wrapper, the sidecar
  binary's shebang, and an example script.
- The `bin/publish` load smoke packs all 28 non-private packages and loads
  every import specifier — each package root plus all exports subpaths, 50 in
  all — under Node, Bun, and Deno, resolving to `dist`; optional peers are
  installed so peer-gated subpaths import.

Closes INTR-299


alexanderguy:

The whole-branch review is clean — no findings.

Two out-of-scope latent items surfaced in review are tracked separately rather than fixed here: the dist-rewrite resolver's extension coverage (INTR-304), and the publish-metadata guard's hardcoded side-effecting package (INTR-305).

The launcher lint guard deliberately does not match a `bun` invocation behind an env-assignment / `command` / `time` prefix, guards only the first `.ts` on a line, and does not see `export * as ns from` — each a documented scope boundary for a form the repo does not use.

PR #90, opened 2026-07-15
Add a launcher wrapper for the guarded publish path

alexanderguy:
## Summary

- Adds a `bin/publish` launcher wrapper so the guarded publish path has the same `bin/<name>` entry point the other launchers — and `docs/PUBLISHING.md` — already assume. It forwards to `bin/publish.ts` and carries no `--conditions=intx-src`, since the publish path resolves no `@intx/*` of its own.
- Extracts the publish gate's tested surface — the target model, `readTargets`, and the version, dependency-expression, ordering, load-matrix, and packed-manifest checks — into `bin/lib/publish-targets.ts`, so the new extensionless wrapper does not shadow `bin/publish.ts` for the test's `./publish` import. The entry point and its test both import from the new module.

## Verification

- `make test` passes: 4060 + 502 tests, 0 failures (live-DB suites included).
- `make lint` passes (via the pre-commit hook): prettier, eslint, gen-api-docs, check-deps, `check-launchers`, exports-shape, publish-metadata.
- `tsc --noEmit -p bin/tsconfig.json` is clean.
- `check-launchers` reports ok with `bin/publish` now among 18 launchers — the basename collision this branch is designed around is resolved.

### DIFF
--- a/bin/lib/publish-targets.ts
+++ b/bin/lib/publish-targets.ts
@@ -1,0 +1,170 @@
+// The testable core of the guarded publish path: the target model and the
+// pure pre-flight checks (version sync, dependency expressions, leaf-first
+// order, the load matrix, and the packed-manifest rewrite), separate from
+// the effectful orchestration that builds, packs, and uploads.
+//
+// This lives in `bin/lib` rather than in the entry point because the entry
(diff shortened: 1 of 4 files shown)


alexanderguy:

PR #92, opened 2026-07-15
Add continuous integration for pull requests

alexanderguy:
## Summary

- Runs the full `make all` pipeline (lint, type check, admin-ui bundle,
  tests) on every pull request and gates merge on it.
- Provisions a Postgres service and wires the database environment so the
  DB and hub-api suites run instead of skipping; a pre-flight assertion
  fails the job if that environment is missing, so a silent skip cannot
  pass as green.
- Bootstraps the database through the repository's own `bin/db-reset`,
  reusing its migrate and grant steps rather than duplicating them.

## Verification

- `make lint` exits 0 on the branch (prettier, eslint, API-docs freshness,
  dependency and launcher checks).
- The full `make all`, including the DB and hub-api suites, is exercised by
  this workflow itself; the first run on this PR is its live verification.

Closes INTR-306

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,98 @@
+name: Continuous Integration
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+concurrency:
+  group: ci-${{ github.event.pull_request.number }}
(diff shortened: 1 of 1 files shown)


alexanderguy:

PR #94, opened 2026-07-16
Recover scheduler timers from the committed object store

alexanderguy:
## Summary

- The workflow-host scheduler's start-time recovery and fire-time deployment attribution read committed git state through the repo store rather than the materialized working tree, so a filesystem that lags the committed object store can no longer make recovery miss a committed `TimerFired` and re-queue an already-fired one-shot timer.
- The repo store exposes `openCommittedReads` (by ref) and `openCommittedReadsAtCommit` (by commit): cache-backed, authorize-gated reads of the committed tree. The workflow-event subscription reads through the by-commit primitive instead of walking the tree with its own git calls.
- The workflow-run event-log filename shape and `runs/`/`events/` prefixes live in one place (the kind handler), and every event-log reader rejects an illegal `<seq>.json` name rather than silently dropping the blob.

## Verification

`make all` exits 0: build, lint, 4080 unit tests, and 361 integration tests all pass. A regression test wipes the materialized `runs/` working tree while leaving the object store intact and asserts recovery still excludes the committed `TimerFired` and still queues the co-resident unfired timer — the lag scenario, which fails on the pre-fix working-tree read.

Closes INTR-308

### DIFF
--- a/apps/sidecar/src/workflow-run-pack-client.ts
+++ b/apps/sidecar/src/workflow-run-pack-client.ts
@@ -685,6 +685,8 @@ export function createWorkflowRunPackPushingRepoStore(
     resolveHead: underlying.resolveHead.bind(underlying),
     getRepoDir: underlying.getRepoDir.bind(underlying),
     openCommittedReads: underlying.openCommittedReads.bind(underlying),
+    openCommittedReadsAtCommit:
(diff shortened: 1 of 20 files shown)


alexanderguy:

PR #95, opened 2026-07-16
Run a real per-step agent for map fan-out

alexanderguy:
## Summary

- A top-level `map` fan-out runs a real per-step agent for each item. A
  scoped iteration step id `<base>[<index>]` resolves to its base step for
  the inference-source, tool-deploy-tree, and grant lookups, so every
  iteration shares the base step's staged assets; per-iteration scratch,
  blob storage, and run-log events stay keyed by the scoped id.
- The scoped-step-id format has a single owner: `scopedStepId` /
  `baseStepId` in `@intx/workflow`, replacing five hand-rolled mint sites
  and two divergent strip regexes.
- The child-runtime step invoker fails loud with
  `ChildStepNotImplementedError` instead of fabricating `{ reply, turn: null }`,
  so a `childWorkflow` fan-out no longer reports a completed run whose agent
  never ran.
- `docs/unified-execution-host-design.md` reflects the built state: `map`
  fan-out reads as real per-item execution, and `childWorkflow`
  child-definition per-step execution reads as the one remaining fail-loud gap.

## Verification

- `make all` passes on HEAD: 4106 fast-suite tests and 363 integration tests
  exit 0 (lint, type-check, and admin-ui bundle included).
- A spawned-child integration test deploys a top-level `map` through the real
  hub and sidecar subprocess and asserts each iteration commits real agent
  output, not the placeholder, with one inference call per item.
- The `childWorkflow` integration suite asserts the loud failure end to end
  while keeping its spawn, sub-namespace, and recursion coverage.

Closes INTR-287


alexanderguy:

alexanderguy:
Follow-on commit adds end-to-end tool coverage: a second case in the map fan-out integration test pins a tool package and asserts each iteration exposes the tool in its reply, driving the tool-deploy-tree base resolution through the real sidecar (previously covered only by unit tests). Verified as a targeted guard — reverting the tool-tree base normalization makes only this case fail. Self-review of the commit returned clean.

PR #96, opened 2026-07-16
Enforce credential-use authorization at agent launch and rotation

alexanderguy:
## Summary

Agent launch and credential rotation resolve provider credentials into an
agent's model `sources[]`. Before this branch, that path shipped the credential
secret into the agent without checking that the agent's creator was authorized
to use the credential: the launch handler ran credential resolution and grant
evaluation as two independent loops that shared no data, and the rotation push
re-resolved secrets the same way. Any principal able to author or deploy an
agent definition could cause a tenant-sourced provider secret to be shipped
into an agent regardless of authorization — a confused-deputy / privilege
escalation. `docs/CREDENTIALS.md` already described this check as enforced; it
was not.

This branch enforces the invariant in the layer that owns it. A provider secret
enters an agent's model sources only if the agent's **creator** holds a
`credential:{id}` / `use` grant for that credential. The check lives in the
model-source resolution layer (`buildSource` / `resolveModelSources`), so both
the launch path (`hub-api` instances route) and the rotation/reconnect push
(`resolveInstanceModelSources`) are gated by one fail-closed check — any grant
effect other than `allow` withholds the secret and surfaces a
`credential_unauthorized` skip (a `not_launchable` 409 on launch; a silent,
never-rejecting no-op on rotation).

## What the branch does

- **Provisioning.** Creating a personal credential mints a durable
  `credential:{id}` / `use` grant (origin `creator`) for the owner in the same
  transaction. Organizational credentials mint none and remain governed by
  role-based access.
- **Backfill.** An idempotent data migration grants existing personal
  credentials' owners the same `use` grant, so fail-closed enforcement does not
  lock out credentials that predate provisioning.
- **Enforcement.** The resolution layer authorizes the creator against
  `credential:{id}` / `use` before a secret joins `sources[]`, fail closed, on
  both the launch and rotation paths.
- **Tenant inheritance.** Because credential resolution reaches a credential
  through the tenant ancestor chain, the credential-use check collects the
  creator's grants across that same chain (via a dedicated chain-aware
  collector), so an inherited credential authorizes correctly. The single-tenant
  grant collection used by general RBAC is unchanged.
- **Docs.** `docs/CREDENTIALS.md` now names the creator (not the freshly minted
  instance principal) as the authorizing subject, matching the enforcement.

## Verification

- `make all` is green on the tip: type-check, lint (Prettier, ESLint, API-docs
  freshness, dependency/launcher/exports checks), and the full test suite —
  4069 fast-suite and 518 integration tests pass, 0 fail, against a real
  Postgres.
- The enforcement is proven by mutation testing: reverting only the enforcement
  source (keeping the tests) makes the reproduction tests fail — the secret
  flows through for a non-owner creator on both launch and rotation — and
  restoring the fix makes them pass. The regression tests use a **non-owner**
  creator, since a tenant owner's wildcard grant would mask the vulnerability.
- New tests pin the authorization invariants (owner and `credential:*`/`use`
  satisfy the check; admin/member defaults do not; `ask`/`deny`/no-match fail
  closed) and cover the multi-tenant inheritance cases (inherited credential
  authorizes; child-tenant owner still authorizes; a creator with no grant
  anywhere in the chain is withheld).

Closes INTR-197.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -279,6 +279,7 @@
       "name": "@intx/db",
       "version": "0.2.2",
       "dependencies": {
+        "@intx/authz": "workspace:*",
         "@intx/log": "workspace:*",
         "@intx/types": "workspace:*",
         "arktype": "catalog:",
--- a/docs/CREDENTIALS.md
+++ b/docs/CREDENTIALS.md
@@ -93,6 +93,12 @@ The grant validation ensures that a definition creator cannot grant an agent acc
 
 At launch time, if the agent requires invoker-sourced credentials, the invoker's grants are additionally validated. The effective credential set is the union of tenant, creator, and invoker credentials, subject to each party's authorization.
 
+### Model-Source Credential Authorization (Interim)
+
+Model-source (tenant-catalog) credentials carry an additional gate: the secret behind a catalog offering only enters a launchable inference source when the agent's creator holds a `credential:{id}` / `use` grant for the referenced credential. This is enforced fail-closed — anything other than an `allow` effect (including `ask`, `deny`, or no matching grant) withholds the secret — at agent launch and again on credential rotation, so a rotated secret is never pushed to a running instance whose creator lacks the grant.
+
+This is a deliberate interim tightening. Catalog credentials are tenant-owned (tenant-source), and under the source-based authority model (tenant/creator/invoker) a tenant-source credential would be authorized by tenant/role policy rather than by a per-principal `use` grant. Gating catalog-credential use on the creator's explicit grant is stricter than that model requires. Unifying catalog-credential authorization with the source-based model is planned; until then the creator `use` grant is the authoritative check on this path.
+
 ## Walk-up Resolution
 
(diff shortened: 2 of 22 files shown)


alexanderguy:
Update since open: the branch now labels the creator-`use` gate on model-source (tenant-catalog) credentials as a deliberate **interim** tightening — documented in `CREDENTIALS.md` and at the enforcement site — rather than the final authority model. Catalog credentials are tenant-owned; under the source-based authority model a tenant-source credential would be authorized by tenant/role policy rather than a per-principal grant, so the creator-`use` gate is stricter than that model requires (safe and fail-closed, but not the end state).

Also added a test proving an invoker's source override cannot bypass the gate: pinning a provider whose credential the creator is not authorized for withholds the secret and makes the model unavailable, so an invoker cannot use an override to spend an unauthorized credential.

Unifying catalog-credential authorization with the source-based model (and the deferred conditional-grant / ask-deny resolution-layer tests) is tracked in INTR-311.

PR #97, opened 2026-07-18
Enable agents to suspend on external signals and resume durably

alexanderguy:
## What this does

Agents can suspend mid-run when they need an answer from outside and resume durably when it arrives, surviving restarts. This builds the signal substrate and wires the first flow — human approval — onto it, tracer-first.

Before this, a tool call that needed approval had nowhere to go: the approval was rejected, the model saw an error, and no durable record of a pending request existed. Now:

- **A tool call that needs approval suspends the agent** instead of erroring. The parked call, its correlation, and the pending approval are recorded durably (an `approval` row and a `signal_correlation` row, co-written at the hub).
- **The run parks on a reserved `__signal__:<correlationId>` channel and survives restarts** — the park lives in the durable log, and a crashed suspended step re-parks on resume against its original attempt and deadline.
- **An operator resolves the request through the hub HTTP route.** The resolver authorizes the approver, claims the correlation as terminal in one transaction, and delivers the decision to the parked run. The claim is single-shot (`resolved_at IS NULL`) with a persisted `signalId`, so a duplicate approval, webhook retry, or double-click cannot resume a run twice.
- **On approval the original tool actually runs.** The reactor re-dispatches the parked call through a one-shot gate bypass keyed on the call id — so the *authorized* call is the one that executes — injects the real result into history, and the agent re-infers exactly once. On rejection or timeout the parked call is answered with a synthesized error tool_result and the agent continues; no dangling tool_use, no re-park loop.
- **Approving "once" applies to this call only.**

The resumption machinery is generic: a single `(kind, outcome)` taxonomy drives how a parked run resumes, so the remaining flows are wiring rather than new mechanism. `approval` is the only arm wired here.

## How it's verified

- `make all` is green on the branch head: build, lint, `tsc -b`, and both test suites — the fast unit suite and the integration suite, which exercises the real Postgres migrations (0038→0041) and the real hub + sidecar stack.
- An end-to-end capstone integration test drives the whole round trip through the real stack with no stubbed suspend or resume: `ask` → suspend → durable park → register → approve via the real HTTP route → resolve → resume → the approved tool runs exactly once → the run completes with the real result. Its mock re-issues the call until it sees a real tool_result, so it loops (times out) against the pre-fix code and passes only when the tool genuinely runs; a fast companion test pins that discrimination so the integration test cannot silently go vacuous.
- Each commit was reviewed for approach and critiqued after landing; the branch was then reviewed as a whole for cross-commit invariants (migration composition, enum/type drift, dead code, message coherence).

## Not included (follow-ups)

- The other six flows (payment, credential refresh, peer-agent offering, inbox reply, fork completion) reuse these rails — deferred by design.
- `scope: "always"` (a standing grant) and approver pre-authorization of a not-yet-reached call both need the tool identity captured at suspend time; the "always" path fails closed with a 400 until then. Tracked in INTR-331 / INTR-334.
- The async-tool pending-marker path currently shares the `approval` kind as an interim; giving it its own `SignalKind` so the resume switch keys on kind is tracked in INTR-335.
- Redelivery of a recorded decision on a failed post-commit push is tracked in INTR-333.


alexanderguy (commented):
## Whole-branch review (reviewer-of-record)

Reviewed `origin/main` (`6927e7e4`) .. `6820d75d` — 18 commits. This pass covers
the whole-branch / cross-commit invariants and the reviewer-of-record checks.

### Reviewer-of-record checks
- `git diff <base>...HEAD --stat` — no `Bin` markers on any source/text file;
  changed files are confined to the signal-substrate subsystems (types, db,
  inference, workflow, workflow-host, hub-*, sidecar, storage, tests, docs).
- `git log --oneline <base>..HEAD` — 18 commits, all on-topic.
- Commit messages — no subject over 72 chars, no body line over 72 chars, no
  external tracker references, no cross-commit or PR references in any body.

### Cross-commit invariants (four subsystem passes — clean)
- **DB / migrations** — the migrations compose from a fresh DB; the final schema
  matches the drizzle source and the latest snapshot; stores, `parse-row`, and
  enums agree with the final columns; the `claimTerminal` single-shot idempotency
  (`resolved_at IS NULL` + persisted `signalId`) is coherent.
- **Inference / reactor** — the `(kind, outcome)` resume switch is total
  (`assertNever`); every resume event has both a producer and a consumer, with no
  unreachable branches; `pendingToolResults` is seeded and decremented
  consistently across the rails, and the reject/timeout `resume.tool_result` path
  leaves it untouched; boundary parsing of a delivered decision fails loud.
- **Workflow / agent** — `agent.send`'s suspended outcome is handled at every call
  site; the `@intx/types` dependency is used; crash-resume reopens the store keyed
  by the same reduced-state attempt the recovery path reads.
- **Types / hub / sidecar** — the sidecar→hub register path is field-consistent
  end to end; the approval HTTP route matches the store signatures; docs match the
  implemented behavior; comments and user-facing text carry no external references.

### Deferred (tracked)
- INTR-335 — async-tool pending markers share the `approval` kind, discriminated
  by `suspendedCall` presence as an interim.
- INTR-331 — the approval snapshot columns (`toolDefinition`, `toolArguments`) are
  nullable pending the `ToolDefinition` threading through the suspend path.

### Gate
`make all` exits 0 on `6820d75d`: build, lint, `tsc -b`, and both suites — the
fast unit suite and the integration suite (real Postgres migrations + real hub +
sidecar, 538 integration tests).

Verdict: clean. Cross-commit invariants hold and the tree is fully gated.

PR #98, opened 2026-07-18
Enrich the approval row with the suspended tool snapshot

alexanderguy:
## Summary

- Build an approver-facing tool snapshot (name, description, input
  schema, arguments) at the authz ask branch and thread it end to end
  to the hub, which co-writes it onto the approval row.
- Require an approval park to carry its snapshot: the park is a
  discriminated `approval` shape whose snapshot is mandatory, and the
  runtime throws at the park emit site if a correlated suspend arrives
  without one, so the invariant fails loud at the layer that owns it.
- Tighten the approval row's snapshot columns to non-null, now that the
  ask rail is the only co-writer and always carries a snapshot.
- Implement the approval list and detail API endpoints and an admin
  panel that renders the snapshot, so an approver can see the tool call
  they are deciding on.

## Verification

- `make all` passes on this branch: lint, whole-graph type check,
  admin-ui production bundle, and the test suite exit 0.
- The snapshot is enforced at every boundary it crosses: the park emit
  throws in-process, the hub-link producer throws, the register frame
  requires it at the wire, and the approval columns are NOT NULL, so no
  approval row is co-written null. The 128 KiB snapshot size cap is
  rejected at both the park.notify and register frame boundaries.
- An in-process end-to-end test threads a real ask suspension's snapshot
  to the park boundary; an allowed call runs the tool without parking or
  manufacturing a snapshot. The cross-process integration test asserts
  the snapshot content reaches the co-written approval row verbatim.
- The NOT NULL migration assumes an empty approval table (the feature
  is unreleased); it sets the columns non-null with no backfill, so it
  must run only where no prior approval rows exist.

Closes INTR-331


alexanderguy:
Self-review (`code-review` skill) over `origin/main..HEAD` returned clean.

PR #99, opened 2026-07-20
Re-register parked approval correlations after a hub or child reconnect

alexanderguy:
## Summary

- When an agent step parks on an approval gate, the sidecar re-registers its
  correlation with the hub whenever a fresh child becomes addressable or the hub
  reconnects, so a run parked while the hub was down at suspend time stays
  approvable.
- The workflow child enumerates its parked approval correlations from durable
  state (both warm single-step and cold per-attempt layouts) and answers the
  supervisor's query over the signed control channel.
- The reactor commits a suspended step's pending operation before signalling
  `reactor.gate.blocked`, and orders that signal after the commit and before any
  racing gate clear, so the approval snapshot is durably persisted and
  reconstructable after a restart.

## Verification

- `make all` passes (4752 tests, 0 failures).
- The reconnect flow works end-to-end against the real deploy harness: a warm
  agent parked on an approval, its hub link dropped and its correlation rows
  deleted, becomes approvable again after reconnect
  (`tests/workflow-deploy/reconnect-reemits-parked-correlation.test.ts`).

Closes INTR-333

### DIFF
--- a/apps/sidecar/src/conversation-state.ts
+++ b/apps/sidecar/src/conversation-state.ts
@@ -946,7 +946,7 @@ function parseJsonOrThrow(raw: string, label: string): unknown {
   }
 }
 
-function isErrnoNotFound(cause: unknown): boolean {
+export function isErrnoNotFound(cause: unknown): boolean {
   if (cause === null || typeof cause !== "object") return false;
   const code = (cause as { code?: unknown }).code;
   return code === "ENOENT";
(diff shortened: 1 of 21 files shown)


alexanderguy:

- Reactor ask-rail deferred-clear replay: the raced ask-rail path (a gate
  timeout firing during the suspend's durable commit, replaying the synthetic
  `resume.tool_result` after `blocked`) has no dedicated regression test. A
  pinning unit test is not constructible — on the ask rail the pre-fix and
  post-fix behaviours differ only in in-memory state and an internal enqueue
  that drains after the held commit in both cases, so the external event
  sequence is identical and any test would pass without the fix. The deferral
  mechanism is pinned by the plain-path race test, and the ask-rail clear branch
  is pinned (non-raced) by an existing timeout test. Greybeard ruling: waive.

Otherwise the whole-branch review returned clean.

PR #100, opened 2026-07-20
Register a control-plane suspension durably across a park-boundary crash

alexanderguy:
## Summary

- Notify the hub of a control-plane park only after the suspension is durable in the workflow log, so a crash between the notify and the flush cannot strand a hub approval row for a correlation the log can no longer reconstruct.
- Recover a step that crashed mid-park (durable `StepStarted`, unflushed `SignalAwaited`) as awaiting-signal instead of failing the run: reconstruct the `SignalAwaited` from the reactor's durable pending approval operation and re-park on the original correlation without re-invoking the agent.
- Acknowledge the suspension registration at the hub once its co-write resolves, and retry it from the sidecar link until acked, so a register frame lost on an open socket is recovered without waiting for a reconnect edge.

## Verification

- `make all` passes: lint, `tsc -b` type-check, the admin-ui bundle, and the full test suite exit clean.
- The reordered notify, the crash-mid-park recovery classifier and its sidecar-side durable readers, and the ack/retry state machine are covered by co-located unit and component tests; the reconnect re-emit resolves a recovered park to completion end-to-end against a real hub, sidecar subprocess, and Postgres.

Closes INTR-336


alexanderguy:

PR #101, opened 2026-07-20
Close a teardown race in signal correlation registration

alexanderguy:
## Summary

- Resolve a workflow deployment and co-write its signal correlation and
  approval inside one transaction that holds a `SELECT ... FOR UPDATE` lock
  on the deployment row, so the `deployed` status check and the co-write are
  atomic against a concurrent deployment teardown.
- Constrain `signal_correlation.deployment_id` with a foreign key to
  `workflow_deployment(id)`, cascading on delete, matching the sibling
  `approval` row's referential guarantee.

## Verification

- The full pipeline passes (`make all` exits 0: lint, type-check, admin-ui
  bundle, tests).
- A deterministic interleaving test flips a deployment off `deployed` while a
  register is in flight and asserts the register blocks on the row lock, then
  throws and writes no rows; it fails without the lock, so it is a genuine
  regression guard.
- The foreign-key migration is safe to apply: the register path is the sole
  writer of `signal_correlation` and always resolves a real deployment, and
  the table is unreleased, so no environment carries rows that could orphan
  the constraint.

Closes INTR-338


alexanderguy:

PR #102, opened 2026-07-21
Paginate the admin approvals list with auto-load on scroll

alexanderguy:
## Summary

- The tenant approvals panel reads the endpoint's keyset cursor through an
  infinite query and auto-loads the next page as the operator scrolls to the
  bottom, so every pending approval is reachable rather than only the first
  page (default 50).
- `approvalsRequestPath` builds each page's request path — no cursor on the
  first page, the opaque cursor URL-encoded on subsequent pages.

## Verification

- `make all` passes: lint, `tsc -b`, the admin-ui `vite build`, and the test
  suites (exit 0).
- `approvalsRequestPath` is unit-tested for the first-page, cursor, and
  URL-reserved-character cases.
- Driven in a browser against a seeded tenant: 50 rows render on first load
  and auto-load to all 55 on scroll; a 5-row tenant stays a single page with
  no extra fetch; a 0-row tenant shows "No pending approvals."

Closes INTR-337

### DIFF
--- a/apps/admin-ui/src/lib/queries/tenants.test.ts
+++ b/apps/admin-ui/src/lib/queries/tenants.test.ts
@@ -1,0 +1,23 @@
+import { describe, test, expect } from "bun:test";
+
+import { approvalsRequestPath } from "./tenants";
+
+describe("approvalsRequestPath", () => {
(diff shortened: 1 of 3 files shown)


alexanderguy:

PR #103, opened 2026-07-21
Paginate the remaining admin-ui lists behind a shared cursor pattern

alexanderguy:
## Summary

- A shared `infiniteListQuery` factory, `usePaginatedList` hook, and
  `PaginatedListSentinel` component replace the hand-rolled infinite query
  and load-more observer that the approvals list carried inline.
- Every remaining tenant list page (principals, roles, agents, instances,
  grants, credentials, wallets, offerings, models, model providers, model
  offerings) and the dashboard's tenant and instance lists consume the
  cursor, so operators reach every row by scrolling instead of hitting a
  silent first-page cap.
- Queries left with no consumer after the migration are removed.

Providers need no change: their only surfaces are dropdowns, and the
inherited-providers endpoint already returns the full set unpaginated.

## Verification

- `make all` passes: lint, strict type-check, the admin-ui production
  bundle, and the full test suite all exit 0.
- `pagination.test.ts` covers the cursor request-path builder, including
  the query-string merge the running-instances list relies on.

Closes INTR-341


alexanderguy:

PR #104, opened 2026-07-21
Add a per-source quirks parameter to provider adapter factories

alexanderguy:
## Summary

- Provider adapter factories and the registry's `resolve` accept an optional
  opaque `quirks` bag, which each factory validates into its own typed arktype
  shape; an absent bag reproduces each adapter's prior universal behavior.
- The openai adapter (which also serves `openai-compatible`) exposes two
  per-source quirks — `forceAssistantReasoningContent` and `reasoningFieldNames`
  — with unknown keys and unreadable field names rejected loudly at construction.
- The anthropic and google-genai adapters declare empty quirk shapes that reject
  any populated bag, so a bag meant for another provider fails at the boundary.

## Verification

The build passes (`make all` exits 0; 4269 + 391 tests). No in-tree caller
behavior changes: the sole production caller resolves adapters with no quirks
and is unchanged. Tests cover both quirks' default and non-default paths, the
exported validators from the package boundary, and quirk resolution through the
builtin registry into a real factory.

Closes INTR-139

### DIFF
--- a/packages/inference/src/adapter.test.ts
+++ b/packages/inference/src/adapter.test.ts
@@ -40,6 +40,38 @@ describe("createAdapterRegistry", () => {
     expect(seen).toEqual([source]);
   });
 
+  test("forwards the quirks bag to the factory untouched", () => {
+    const seen: unknown[] = [];
+    const registry = createAdapterRegistry({
+      test: (_source, quirks) => {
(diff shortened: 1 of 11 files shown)


alexanderguy:

PR #105, opened 2026-07-21
Carry per-deployment adapter quirks through the inference catalog

alexanderguy:
## Summary

- The model offering catalog row carries a nullable `quirks` jsonb bag, the
  runtime `InferenceSource` type carries it, and the catalog resolver hands a
  resolved source's quirks to the adapter factory at instantiation.
- The catalog offering API accepts quirks on create, replaces or clears it on
  patch, and returns it; the development seed provisions the known deployments
  — Anthropic/OpenAI/Gemini direct plus the Fireworks, Moonshot, OpenRouter,
  and OpenCode Zen kimi backends — each with an explicit quirks bag.
- A CI guard validates every seeded offering's quirks bag against its adapter's
  quirk schema, so a bag the adapter would reject cannot be seeded.

## Verification

- `make all` passes: lint, `tsc -b --noEmit --force`, the admin-ui production
  bundle, and the unit and integration test suites all exit 0.
- Migration `0044` applies cleanly against a fresh schema, landing
  `model_offering.quirks` as a nullable `jsonb` column.
- The seed guard test validates each seeded offering's quirks against its
  adapter's quirk schema.
- The seed's catalog rewrite type-checks and lints; a live `bin/dev --seed`
  run is recommended as manual QA, since it resets the development database.

Closes INTR-140

### DIFF
--- a/bin/lib/catalog-seed-data.test.ts
+++ b/bin/lib/catalog-seed-data.test.ts
@@ -1,0 +1,41 @@
+import { describe, test, expect } from "bun:test";
+import { type, type Type } from "arktype";
+
+import {
+  AnthropicQuirks,
+  GoogleGenAIQuirks,
+  OpenAIQuirks,
+} from "@intx/inference/providers";
+
+import { catalogProviders, type CatalogPlugin } from "./catalog-seed-data";
+
+// Each catalog plugin maps to the adapter quirk validator that governs the
+// shape its offerings may carry. `openai` and `openai-compatible` share the
+// OpenAI adapter, so they share its validator.
+const quirkValidatorByPlugin: Record<CatalogPlugin, Type> = {
+  anthropic: AnthropicQuirks,
+  openai: OpenAIQuirks,
+  "openai-compatible": OpenAIQuirks,
+  "google-genai": GoogleGenAIQuirks,
+};
+
+function isPlainObject(value: unknown): boolean {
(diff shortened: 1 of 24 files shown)


alexanderguy:

A whole-branch review ran against `origin/main`. One finding — silent
`continue` guards in the seed's catalog loop — is fixed (they now fail loudly
via `fatalMissing`). The branch carries three further items the reviewer
raised, each reviewed by Greybeard and ruled consistent with established
codebase conventions:

- `packages/hub-api/src/routes/model-offerings.ts`: the offerings-list
  formatter parses each row, so one corrupt stored row fails the list loudly
  rather than being served. Authorized by Greybeard: this mirrors the existing
  `formatApproval` list pattern, and a corrupt row is unreachable through the
  validating write path.
- `packages/types/src/catalog.ts`, `packages/db/src/parse-row.ts`: `quirks` is
  validated as the shared `Record<string, unknown>` jsonb shape, which accepts
  arrays; a non-empty array is rejected loudly at the adapter factory.
  Authorized by Greybeard: consistency with the codebase-wide jsonb convention.
- `bin/lib/catalog-seed-data.test.ts`: the seed guard's `isPlainObject` is
  stricter than the boundary validators. Authorized by Greybeard: it is the
  load-bearing plain-object check for seed data, not redundant.

PR #106, opened 2026-07-21
Seed catalog offering capabilities from the discovery support matrix

alexanderguy:
## Summary

- Home the shared model-capability vocabulary in `@intx/types` (the wire
  capabilities the runtime supports, plus curated `long-context` /
  `prompt-caching`); `@intx/inference-discovery` imports and extends it, so
  production code never depends on the discovery package.
- `catalogCapabilitiesFor` expands a `(provider, model)` support-matrix query
  into the catalog capability set, lighting up a buffered base from a proven
  streaming variant and projecting onto the production vocabulary.
- The catalog seed derives each offering's wire capabilities from the matrix
  through that helper; only the curated tags the matrix cannot prove are
  hand-set. Offerings whose exact tuple is not yet probed carry empty wire sets
  (rerun tracked in INTR-343).
- A colocated CI guard keeps seeded capabilities backed by the matrix and
  forbids hand-authoring a wire capability into the curated list.
- `docs/ARCHITECTURE.md` and the discovery README document the one-way
  matrix→catalog seed direction.

## Verification

- `make all` passes (lint, `tsc -b --force`, admin-ui bundle, tests), exit 0.
- The expansion helper, the seed derivation, and the CI guard are covered by
  unit tests; the seeded demo agent resolves against its offering's curated
  `long-context`.
- Runtime support for serving a buffered response when only the streaming
  variant is proven is tracked in INTR-342.

Closes INTR-142


alexanderguy:

Self-review plus a high-effort whole-branch review. The two actionable
findings — a `string[]` widening in the seed capability helper and a type
assertion in the guard — are fixed. The branch carries these intentional
exceptions, confirmed with the issue author:

- Offerings whose exact `(provider, model)` tuple is not in the discovery
  matrix advertise empty wire capability sets rather than borrowing another
  model's evidence; rerunning discovery for them is tracked in INTR-343.
- The vocabulary change ships no data migration: the offering capabilities
  column is populated only by the seed and reset from it under the prototype
  policy, so no durable row carries the old vocabulary.
- The admin-UI offering picker offers the full capability vocabulary; the
  runtime does not guard capability writes (the matrix guard is a CI check
  over the seed data).

The streaming-superset-by-suffix expansion and the guard's non-empty
source-live check were reviewed and endorsed as the intended design.

PR #107, opened 2026-07-22
Stop the openai adapter forcing reasoning_content by default

alexanderguy:
## Summary

- The openai adapter defaults `forceAssistantReasoningContent` to false:
  `reasoning_content` is emitted on an assistant message only when that turn
  carries thinking. Forcing it onto every assistant message requires the
  explicit per-source quirk.
- The six catalog offerings whose backends require the forced field carry
  `forceAssistantReasoningContent` explicitly, so the strict default leaves
  them unchanged while any source without the quirk gets the strict protocol.

## Verification

The build passes (`make all` exits 0: lint, `tsc -b`, the admin-ui bundle, and
both test passes). The openai adapter suite pins the strict default (no
`reasoning_content` on a thinking-free assistant turn), the explicit-quirk
forcing path, and the reasoning-field precedence.

Closes INTR-141


alexanderguy:

PR #108, opened 2026-07-22
Relocate discovery-probe fixtures and docs into their provider packages

alexanderguy:
## Summary

- Each provider's captured wire corpus lives inside the discovery package
  that probes it (anthropic, google-genai, and opencode-zen → the
  openai-compatible package), next to that provider's plug-in.
- `getFixtureDir` resolves a per-provider fixture root, so the probe runner
  and the replay suite follow the corpus with no hardcoded single wire root;
  a fixture-bearing entry with an unmapped provider fails loudly.
- The Anthropic and OpenCode Zen observed-vs-documented narratives move into
  their provider packages' `docs/`, alongside the fixtures they describe.

## Verification

The build passes (`make all` exits 0: lint, `tsc -b`, the admin-ui bundle,
and both test passes — 4330 + 394 tests, 0 failures). The replay corpus, the
per-provider adapter fixture tests, and the structured-output round-trips all
read from the new homes.

Closes INTR-345


alexanderguy:

PR #110, opened 2026-07-22
Make each workflow run a first-class authorization principal

alexanderguy:
## Summary

- Each workflow run is its own authorization principal (`kind:"workflow"`, `refId=runId`), minted at the run's birth, with per-run grants materialized at the trigger route and read behind a dispatch barrier.
- Run grants inherit uniformly: a spawned child inherits the parent run's grants (multi-hop); a mail-triggered run gets the receiving definition's creator-resolved grants; a run that cannot resolve its grants fails closed — no silent empty-grant fallback.
- A `workflow_run` table anchors approvals and signal-correlations to the run via foreign key; a terminal seam flips run status and deactivates the run principal; pinned tools authorize via a sidecar tool-mark floor that holds `ask` under grant precedence.

## Verification

- The full pipeline passes (`make all`: build, lint, and both test passes green, including the real-Postgres `tests/db/` suite).
- Mail-triggered materialization is covered by a DB-backed test (real deployment id; idempotent redelivery; rejection writes no orphan).
- Approval-gating is proven against the real substrate: a pinned `ask` tool suspends; an ungated tool runs through.

Closes INTR-339


alexanderguy:

PR #114, opened 2026-07-28
Run CI on pushes to main, not only pull requests

alexanderguy:
## Summary

- Trigger `make all` on pushes to `main`, not only on pull requests, so every
  commit that lands on `main` is validated — including one that reaches `main`
  outside the PR flow.
- Key the push concurrency group on the commit SHA, so each main commit runs in
  its own group and none is evicted from a shared single pending slot. Pull
  requests still group by number and cancel superseded runs.

## Verification

- `make lint` and `make build` pass on the branch (exit 0).
- The full Postgres-backed `make all` runs in CI on this PR.

Closes INTR-381

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,10 +3,24 @@ name: Continuous Integration
 on:
   pull_request:
     types: [opened, synchronize, reopened]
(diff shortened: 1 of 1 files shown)


alexanderguy:

PR #116, opened 2026-07-28
Fail over on quota exhaustion instead of retrying the same source

alexanderguy:
## Summary

- The reactor fails over to the next inference source on `quota_exhausted`
  instead of running a redundant same-source backoff; the `runInference`
  wrapper is the sole owner of mechanical quota retry.
- Worst-case HTTP calls for a sustained `quota_exhausted` are 3 per source
  (the wrapper's cap), with no reactor-level multiplier on top.
- `docs/INFERENCE.md` and `docs/ARCHITECTURE.md` describe the wrapper-owned
  contract and the single-source consequence: no long same-source wait, with
  a custom `RetryPolicy` as the escape hatch for extended pacing.

## Verification

- `make all` passes (exits 0, 0 test failures) on the branch rebased onto
  `origin/main`.
- The `createReactor — source failover` suite asserts `quota_exhausted` now
  fails over immediately (`["s0","s1"]`); wrapper-level retry behavior stays
  covered by `tests/inference/retry.test.ts`.

Closes INTR-130


alexanderguy:

PR #118, opened 2026-07-29
Add a Playwright harness that brings the full stack up headless

alexanderguy:
## Summary

- A Playwright suite (`tests/admin-ui-e2e`) provisions a fresh database per
  run, spawns the hub on a dynamic port, serves the built admin UI through
  vite preview behind a same-origin `/api` proxy, and drives a real Chromium
  browser through sign-up to an authenticated page before tearing everything
  down.
- `@intx/test-harness/db-provision` creates, migrates, grants, and
  force-drops uniquely named databases over the maintenance connection
  `bin/db-reset` already relies on; the migration role gains no new
  privileges.
- `make test-e2e` runs the suite outside the default build and unit passes;
  continuous integration caches the Chromium install and reuses the bundle
  from the preceding `make all` step.

## Verification

- `make all` exits 0 (lint, forced type check, admin UI bundle, both test
  passes).
- `make test-e2e` passes: a real Chromium sign-up flows through the preview
  proxy's rewritten Origin, the session cookie persists, and the
  authenticated dashboard renders; teardown leaves no stray database.
- The provisioner carries its own integration test proving the hub-role
  grants via a hub-role data round-trip.

Closes INTR-390

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,6 +67,18 @@ jobs:
       - name: Install dependencies
         run: bun install --frozen-lockfile
 
+      # Cache the Playwright browser download, keyed on the resolved
+      # @playwright/test version that bun.lock pins. On a hit the install
+      # step below is a no-op; on a miss it re-downloads Chromium.
+      - name: Cache Playwright browsers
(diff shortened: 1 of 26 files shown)


alexanderguy:

PR #121, opened 2026-08-03
Validate publishable sideEffects per-manifest and document dist-rewrite

alexanderguy:
## Summary

- Derive each publishable package's `sideEffects` from its own manifest,
  validated present, well-formed, and naming at least one shipped
  module, instead of hardcoding `@intx/log`. `--fix` seeds `false` only
  when absent and never overwrites a declaration.
- Validate each declared `sideEffects` glob against what the package
  ships: it must match a file on disk or fall under the package's
  `files` allowlist, so a typo or a source-only declaration is caught at
  lint.
- Document why `bin/dist-rewrite.ts`'s resolver probes only
  `.js`/`index.js` — extensioned specifiers are already resolved, and an
  extensionless non-`.js` target cannot be emitted under
  `moduleResolution: "bundler"` — with a regression test that a
  compiling `./data.json` import survives the rewrite untouched.

## Verification

- `make all` passes (lint, type-check, admin-ui bundle, tests) on the
  branch.
- The publish-metadata guard passes against all 28 non-private packages;
  `@intx/log` passes on its own src+dist `sideEffects` declaration.

Closes INTR-303
Closes INTR-305

### DIFF
--- a/bin/dist-rewrite.test.ts
+++ b/bin/dist-rewrite.test.ts
@@ -162,6 +162,26 @@ test("a sibling file resolves to `.js` and a directory barrel to `/index.js`", (
   expect(out).toContain('import { b } from "./bar/index.js";');
 });
 
+test("a compiling `.json` import resolves without aborting", () => {
+  // A `resolveJsonModule` import that type-checks emits with its `.json`
+  // extension, which `isRewritable` treats as already resolved. The
+  // rewriter must leave it in place, not report it unresolved. A naive
+  // "make `.json` rewritable" change would send `./data.json` to the
+  // resolver, which finds neither `./data.json.js` nor
+  // `./data.json/index.js`, falsely report it unresolved, and abort the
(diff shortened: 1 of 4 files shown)


alexanderguy:

Whole-branch review returned clean after fixes.

Two intentional present-state properties of the branch, both reviewed
and operator-accepted:

- The guard validates that each publishable package's `sideEffects` is
  present, well-formed, and names at least one shipped module — but it
  does not re-establish a package-specific "must not be `false`" rule.
  Whether a module has an import-time side effect is a fact owned by the
  package's own manifest, not a name registry in the guard.
- `--fix` seeds `false` only when `sideEffects` is absent and fails
  loudly on a malformed value rather than normalizing it, since there is
  no longer a canonical value to normalize to.

Per-filename verification of emitted `dist` globs needs the built tree,
so it is deferred to INTR-415 (the lint guard enforces the array-level
"at least one shipped entry" floor in the meantime).

PR #122, opened 2026-08-04
Verify sideEffects globs resolve in the built tree before publishing

alexanderguy:
## Summary

- `checkBuiltSideEffects` confirms every non-`false` `sideEffects` glob in each
  publishable package resolves to a real file in the built package directory
  (after `build-dist` emits `dist/`), with no `files`-coverage escape — so an
  emitted-path typo like `./dist/registr.js` for `./dist/register.js` is caught
  instead of being silently tree-shaken by a consumer's bundler.
- The guarded publish path runs the check right after `build-dist`, aborting the
  run (dry run and `--execute` alike) when a declared glob names a file the build
  never produced.

## Verification

The build passes (`make all` exits 0; 4803 + 423 tests). Unit tests cover a glob
matching an emitted file, an emitted-path typo, a mixed array, the `@intx/log`
src-and-dist shape, the `false` skip, and the absent/malformed precondition throw.

Closes INTR-415

### DIFF
--- a/bin/publish-metadata.test.ts
+++ b/bin/publish-metadata.test.ts
@@ -10,6 +10,7 @@ import { tmpdir } from "node:os";
 import { dirname, join } from "node:path";
 
 import {
+  checkBuiltSideEffects,
   checkWorkspaceDescriptions,
   checkWorkspaceMetadata,
   expectedFiles,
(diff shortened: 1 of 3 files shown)


alexanderguy:

PR #123, opened 2026-08-05
INTR 372 - Allocate provision and route isolated workflows

FarinM:
## Summary

This PR adds exclusive sidecar placement for workflows. Shared capacity remains the default, but workflows or tenant policy can now require a dedicated sidecar that accepts only that deployment. Exclusive deployments are prepared and managed durably by the Hub. Deployment, trigger, signal, approval, and workflow-state traffic is fenced to the allocation's currently authenticated generation.

## What this adds

- A versioned sidecar provisioner contract and plugin registry. The Hub does not include a built-in infrastructure backend; exclusive deployment fails closed when no provisioner is configured.
- Durable, secret-free workflow launch specifications used to reconstruct deployments.
- A persisted sidecar allocation lifecycle with reconciliation leases, retry scheduling, generation fencing, reconnect grace, and idempotent provision/destroy operations.
- Allocation-aware authentication and routing. Allocated workers are excluded from shared scheduling and may register only their assigned deployment.
- Durable mail, signal, and approval dispatch. Messages remain unsettled until accepted workflow Git state proves they were consumed.
- Workflow-run pack fencing so stale, shared, or wrong-generation workers cannot advance an exclusive deployment's state.
- Optional replacement recovery that preserves the deployment ID, address, workflow-run repository, and committed event history across workers.

  > Automatic replacement recovery is disabled by default. Hub-owned state cannot restore arbitrary sidecar filesystem or isolation-container state, so worker loss fails active runs and releases capacity unless recovery is explicitly enabled.

## Supporting changes

- Add migrations for workflow launch specifications, sidecar allocations, and durable workflow dispatch.
- Restore committed workflow history before replaying work on a replacement worker.
- Re-resolve inference sources during recovery so credential rotation and revocation take effect; raw provider secrets are never persisted in launch specifications.
- Forward workflow tool-package pins through both deployment paths.
- Buffer reactor deliveries until startup restoration completes, and allow aborts to interrupt in-flight inference immediately.
- Traverse complete tenant ancestor chains with cycle detection instead of silently truncating deep inheritance.
- Pin database sessions to UTC so JavaScript dates and PostgreSQL retry deadlines represent the same instant.
- Extend API errors and documentation for inactive allocations and unavailable durable dispatch.

## Verification

`make all` passes, including formatting, lint, generated documentation checks, TypeScript validation, the admin UI production bundle, and all test suites.


apps/hub/src/index.ts
```diff
@@ -206,6 +219,84 @@ const sessionService = createSessionService({
...
+  onReady: async (allocation) => {
+    await workflowAllocationService.deployReadyAllocation(allocation);
+    await workflowDispatchService.requeueForReadyAllocation(
+      allocation.anchorRunId,
+    );
+  },
+});
+
+await sidecarAllocationReconciler.initialize();
+sidecarRouter.events.on("sidecar.disconnect", ({ allocated }) => {
```
alexanderguy, line 280: I think there's a subtle gap here — `handleDisconnect` goes through the fire-and-forget emitter, so if `markConnectionLost` hits a transient DB error it just gets logged and dropped. Since that write is what restores `next_attempt_at`, the row can end up stuck as `allocated` with a null `next_attempt_at`, which `claimNextReconcilable` won't pick up again until a restart. Might be worth either retrying in `handleDisconnect` or running `initialize()`'s repair on the ticker so these get swept back up. (Same pattern on the `mail.inbound.acknowledged` listener just below, if we go that route.)

packages/hub-sessions/src/sidecar-allocation/reconciler.ts
```diff
@@ -0,0 +1,583 @@
...
+    const allocation = await allocationStore.claimNextReconcilable({
+      leaseId,
+      leaseDurationMs,
+    });
+    if (allocation === null) return false;
+    try {
+      await reconcile(allocation, leaseId);
+    } catch (error) {
+      logger.error`Allocation ${allocation.id} reconciliation failed: ${error instanceof Error ? error.message : String(error)}`;
+      await allocationStore.parkReconciliation(allocation.id, leaseId);
```
alexanderguy: Heads up — `parkReconciliation` lands `next_attempt_at` at null when there's no `connect_deadline`, which is the case for `pending`/`replacing` rows. So if `reconcile()` throws before we've bound a sidecar, the row can get parked and never reclaimed. Since this is the catch-all for unexpected failures, `scheduleRetry` might be a safer fit here than parking.

packages/hub-api/src/routes/approvals.ts
```diff
@@ -124,16 +134,36 @@ async function resolveApproval(
...
+        id: sidecarAllocation.id,
+        status: sidecarAllocation.status,
+      })
+      .from(sidecarAllocation)
+      .where(eq(sidecarAllocation.anchorRunId, approval.deploymentId))
+      .limit(1)
+      .for("update");
+    if (
+      allocation !== undefined &&
+      !isSidecarAllocationDispatchable(allocation.status)
```
alexanderguy: One thing I noticed: the dispatchability check runs ahead of `claimTerminal`, so a retried decision on an approval that's already resolved now comes back as `deployment_unreachable`/`workflow_dispatch_unavailable` once the allocation is released, where it used to return `already_resolved`. Could we check the claim first? That'd keep resolved approvals reporting as resolved regardless of what's happened to the allocation.

packages/hub-api/src/routes/workflows.ts
```diff
@@ -553,6 +737,73 @@ export function createWorkflowRoutes({
...
+        );
+      }
+
+      const allocationId = deployment.allocationId;
+      if (allocationId !== null) {
+        if (workflowDispatchService === undefined) {
+          return c.json(
+            {
+              error: {
+                code: "workflow_dispatch_unavailable",
```
alexanderguy, line 826: Small inconsistency — this returns `workflow_dispatch_unavailable` as a 502, but the same thing is a 503 at line 1143 and over in `approvals.ts:582`. I'd lean 503 for all three (502 tends to mean a bad upstream response, which isn't quite what's happening here). Looks like the trigger route also doesn't declare this status in its `responses` block, so folding these into one shared response might tidy up both.

packages/hub-sessions/src/hub-session-lookups.ts
```diff
@@ -445,8 +527,41 @@ export function createHubSessionLookups(
...
+        if (reads !== null) {
+          for (const consumed of await listConsumedWorkflowDispatches(reads)) {
+            if (consumed.address !== anchor.address) continue;
+            await workflowRunDispatchStore.settle(
+              anchor.id,
+              consumed.messageId,
+              now,
+            );
+          }
+          for (const signal of await listReceivedWorkflowSignals(reads)) {
```
alexanderguy: This one's just an efficiency thought — `listReceivedWorkflowSignals` reads every run's full history, and then we filter down to the anchor's with `signal.runId !== anchor.address`. Running on every accepted pack against an ever-growing `runs/`, that adds up over a deployment's life. Since we've got the anchor address here, scoping the read to its own run directory would keep it flat.

packages/db/src/workflow-run-dispatch-store.ts
```diff
@@ -0,0 +1,466 @@
...
+      updatedAt: now,
+    })
+    .where(and(...conditions))
+    .returning();
+  return updated === undefined ? null : parseWorkflowRunDispatchRow(updated);
+}
+
+export function createWorkflowRunDispatchStore(db: DBHandle) {
+  return {
+    async enqueue(
```
alexanderguy, line 234: `enqueue` and `enqueueSignal` are pretty much the same insert/re-select/compare shape, and I noticed the equality check has already started to drift between them. Might be a good candidate for a shared `insertOrReconcile` helper before the two diverge further.

packages/db/src/sidecar-allocation-store.ts
```diff
@@ -0,0 +1,885 @@
...
+    async findByAnchorRunId(
+      anchorRunId: string,
+    ): Promise<SidecarAllocation | null> {
+      const row = await db.query.sidecarAllocation.findFirst({
+        where: eq(sidecarAllocation.anchorRunId, anchorRunId),
+      });
+      return row === undefined ? null : parseSidecarAllocationRow(row);
+    },
+
+    async findBySidecarId(
```
alexanderguy: Looks like `findBySidecarId` doesn't have any callers yet — is it staged for something coming, or safe to drop?

packages/db/src/workflow-run-dispatch-store.ts
```diff
@@ -0,0 +1,466 @@
...
+    },
+
+    async findById(id: string): Promise<ParsedDispatch | null> {
+      const row = await db.query.workflowRunDispatch.findFirst({
+        where: eq(workflowRunDispatch.id, id),
+      });
+      return row === undefined ? null : parseWorkflowRunDispatchRow(row);
+    },
+
+    async findByMessage(
```
alexanderguy: Same question here as `findBySidecarId` — couldn't find any callers for `findByMessage`, tests included. Leftover, or is there a use on the way?

alexanderguy (dismissed):
Took a thorough pass over the exclusive-allocation and durable-dispatch work. Overall the design holds up really well — the generation/lease fencing is consistent, the migrations are careful, and the trust boundaries validate where they should. A few things worth sorting before merge, so marking as changes requested.

The ones I'd call blocking are the three correctness items: two allocations that can end up with a null `next_attempt_at` and never get reclaimed until a Hub restart (the swallowed disconnect write and the `parkReconciliation` case — they share a root cause, so a periodic version of `initialize()`'s repair would close both), and the approval retry that now returns the wrong status once its allocation is released. The rest are smaller: a status-code inconsistency, a history rescan that grows with deployment lifetime, some duplication that's already starting to drift, and two unused store methods.

Inline comments below have the details. Nothing here is a design objection — it's all fixable in place.

_(This review was generated by the pull-request-review skill.)_

packages/hub-sessions/src/sidecar-allocation/reconciler.ts
alexanderguy: Following up after the latest push — the new `repairUnscheduledConnections` sweep is a nice fix and fully closes the disconnect case, but I don't think it reaches this one. Both the sweep and `scheduleReconnectIfUnscheduled` are gated on `status === "allocated"`, and this strand happens on `pending`/`replacing` rows, so they get skipped.

The path that's still open: `reconcileNext` claims a `pending` row, then `bindInitialSidecar` hits a transient DB error mid-transaction (serialization failure, statement timeout, a dropped connection) before it sets `connect_deadline`. The transaction rolls back, the row stays `pending` with `connect_deadline` still null, and the error lands in `reconcileNext`'s catch → `parkReconciliation`, which writes `next_attempt_at = NULL`. From there `claimNextReconcilable` can't see it (`NULL <= now()` is never true), the repair sweep skips it (not `allocated`), and no connect event fires since nothing was ever provisioned — so it sits until the next restart runs `initialize()`. Same shape for `replacing` via `bindReplacementSidecar`.

Swapping `parkReconciliation` for `scheduleRetry` in that catch would close it independently of the sweep, since `scheduleRetry` always leaves a concrete `next_attempt_at`. Flagging so it doesn't look like the sweep already covers it.

packages/db/src/sidecar-allocation-store.ts
```diff
@@ -0,0 +1,924 @@
...
+      fallbackNextAttemptAt: Date,
+    ): Promise<boolean> {
+      const fallback = sql.param(
+        fallbackNextAttemptAt,
+        sidecarAllocation.nextAttemptAt,
+      );
+      const [updated] = await db
+        .update(sidecarAllocation)
+        .set({
+          nextAttemptAt: sql`case when ${sidecarAllocation.connectDeadline} is null then ${fallback} else coalesce(${sidecarAllocation.nextAttemptAt}, ${sidecarAllocation.connectDeadline}) end`,
```
alexanderguy: Small thing on this fallback — the CASE only applies the caller's fallback when `connectDeadline` is null. `reconcileNext`'s catch passes `retryAt(MAX)` intending a backed-off retry, but a `provisioning`/`allocated` row has a non-null `connectDeadline` and a `nextAttemptAt` that's already in the past (it had to be `<= now()` to get claimed), so the ELSE branch keeps that past timestamp and drops the backoff — the row is immediately re-claimable. On a persistent throw (e.g. a `markConnectionReady` write failing during a DB incident) that turns into a hot loop, ~100 attempts/tick every second, piling load onto the database during the exact outage that triggered it, instead of the ~30s backoff. Could the ELSE branch floor `nextAttemptAt` at the fallback too (something like `greatest(coalesce(nextAttemptAt, connectDeadline), fallback)`), so allocated/provisioning rows back off as well?

packages/hub-api/src/routes/workflows.ts
```diff
@@ -553,6 +767,126 @@ export function createWorkflowRoutes({
...
+            return "run-not-running" as const;
+          }
+          if (
+            (await lockWorkflowRunState(tx, deploymentId, deploymentId)) !==
+              "running" ||
+            (await lockWorkflowRunState(tx, deploymentId, runId)) !== "running"
+          ) {
+            return "run-not-running" as const;
+          }
+          await workflowDispatchService.enqueueSignal(
```
alexanderguy: On the exclusive path this `enqueueSignal` runs inside the transaction with no try/catch, and `insertOrReconcileDispatch` throws `conflicts with its durable payload` when the same `signalId` comes back with a different payload. With nothing catching it, that surfaces as a 500 and the payload is dropped — where the old `sendSignalDeliver` path treated a reused `signalId` as idempotent. So a client retrying a signal with a reused id and a changed body now gets a 500 instead of a deterministic 4xx. Could we catch the conflict and map it to a 4xx (and declare it in the route's `responses`)? The approval path is fine here — it mints the `signalId` server-side and gates on `claimTerminal` — so this is just the client-facing signal route.

packages/hub-sessions/src/sidecar-allocation/reconciler.ts
alexanderguy: Confirmed fixed on the current head — `parkReconciliation` now takes a `fallbackNextAttemptAt` and writes it (instead of NULL) when `connect_deadline` is null, so a parked pending/replacing row keeps a concrete `next_attempt_at` and gets reclaimed. Resolving this one. (There's a separate, still-open note on the same function about the fallback being discarded for allocated/provisioning rows — that's the new thread on the `parkReconciliation` CASE.)

alexanderguy (changes requested):
Re-reviewed the current head fresh — the eight items from the first round are all resolved and the build is green. Two correctness items are worth fixing before merge; both are left as inline comments on this push:

- `parkReconciliation` (sidecar-allocation-store.ts) discards the caller's backoff fallback for allocated/provisioning rows, so a persistent reconcile failure re-parks at a past timestamp and turns into a ~100-attempts/tick retry storm against the database during the exact incident that triggered it.
- The exclusive signal route (workflows.ts) lets `insertOrReconcileDispatch`'s payload-conflict throw escape the transaction as an unhandled 500 and drops the payload, where a reused signalId used to be idempotent.

Two lower-severity items are tracked as follow-ups rather than blockers: INTR-441 (federated inbound mail isn't durable across a Hub restart — a pre-existing gap, not a regression introduced here) and INTR-442 (shared mail-trigger can accept a trigger into an already-terminal run and return a false 202). Happy to re-review once the two above are addressed.

alexanderguy (approved):
Approving. The two items from my earlier changes-requested review — the parkReconciliation retry storm and the signal-route 500 on payload conflict — are both fixed on this head and verified, and a fresh from-scratch review plus two adversarial validation passes turned up no correctness blockers.

Some non-blocking follow-ups are being tracked in Linear rather than gating this PR:

- INTR-441 — federated inbound mail to exclusive deployments isn't durable across a Hub restart (pre-existing gap, not a regression here)
- INTR-442 — shared mail-trigger can accept a trigger into an already-terminal run (narrow, non-corrupting race)
- INTR-446 — requeueUnsettled doesn't reset attemptCount (bounded, latency-only)
- INTR-447 — dispatch/reconcile cleanups (wake-race latency, helper duplication, an N+1 on the deploy path)

None of these affect correctness on a reachable path. Nice work on the rework.

PR #124, opened 2026-08-05
Unify the wire and session capture corpora onto one session format

alexanderguy:
## Summary

- Recorded provider captures live in a single session-shaped corpus — a
  `session.json` manifest plus per-exchange byte directories and
  reverse-engineered tool dispatches — stored next to the discovery package
  that produced them instead of inside `@intx/inference-testing`.
- The production `@intx/inference` harness decodes non-streaming JSON
  responses (branching on `Content-Type` at the 2xx boundary), and every
  adapter implements a JSON-response parser, so JSON-bodied captures replay
  through the real `runInference`.
- The catalog derives capability/model/provider identity from the session
  manifest rather than the directory path, and one loader drives the
  parser-regression over the whole corpus; strict canonical request-body
  equality lives behind a separate replay entry point.
- The legacy wire corpus, its wire→session converter, and its format surface
  (`getFixtureDir`, `FixtureManifest`, the wire-rooted fixture-root constant)
  are removed.

## Verification

- `make all` passes (format, lint, typecheck, repo checks, and the full test
  suite).
- The parser-regression replays every fixture-bearing (adapter × capability)
  cell through production `runInference`; one live multi-turn tool session per
  adapter, recorded against real provider endpoints, replays through the strict
  harness.

Closes INTR-143


alexanderguy:

A whole-branch review ran (high effort). Every finding is fixed except two
intentional exceptions, each authorized by Greybeard:

- `packages/types/src/content-type.ts` + `packages/inference/src/harness.ts`:
  `detectResponseKind` rejects a 2xx whose `Content-Type` is neither
  `text/event-stream` nor `application/json` instead of feeding unknown bytes
  to the SSE parser. Deliberate boundary validation — the recorder routes
  through the same function, so no non-conforming provider can enter the
  corpus, and failing loudly beats emitting a silent empty turn.
- `packages/inference-testing/src/session-replay.ts`: replay serves the
  captured `Content-Type` verbatim rather than reconciling it against the
  on-disk body kind. Verbatim is correct fidelity — where a provider mislabels
  a response, production mis-decodes identically, so replay reproduces
  production rather than being more lenient.

Two follow-ups are tracked: INTR-424 (promote the parser-regression to strict
canonical equality per cell) and INTR-426 (guard the catalog base-URL mirror
against drift).

PR #126, opened 2026-08-07
Source workflow definitions from a remote npm registry, end to end

alexanderguy:
Closes INTR-416.

## Summary

Previously a workflow had to be defined inline. This change allows a workflow to be defined in code, published as an npm package, and installed and run by the hub directly from the registry, without executing the package author's code in a trusted context.

## How it works

1. **Install.** The hub resolves the package and its dependencies from the registry, reading metadata only. No author code is executed at this stage.
2. **Probe.** To determine the workflow's requirements (tools and AI models), the hub evaluates the code once inside an isolated sandbox and receives a plain-data summary. Author code never executes on the hub or the host.
3. **Approve.** The operator reviews the requested capabilities and approves them. The hub records a fingerprint of the approved definition.
4. **Run.** On every load — fresh run, resume, and sub-workflow spawn — the sidecar re-verifies the loaded definition against the hub-approved fingerprint and refuses to run it if it no longer matches. This detects a post-approval swap of the on-disk artifact; it is an integrity check, not a sandbox for the code itself (see Known limitation).

## Changes in this pull request

An earlier change established the end-to-end path. This pull request adds support for onTrigger sections — sub-workflows triggered by an event such as an incoming email.

Previously, a code-sourced workflow's onTrigger section could be deployed but could not run its AI step, because its model configuration was never provisioned. This change provisions each section's model from the approved definition, using the same approval check as the parent workflow. A section without an approved model is rejected rather than run unconfigured.

## Rebase onto main

main recently changed how workflow runs are tracked: a deployment is now represented as its own run record. This branch has been rebased onto that model and adapted accordingly, and it adds one database migration.

## Known limitation

A workflow may reference a separate, independently approved workflow. In that case the hub cannot re-verify the referenced workflow against a fingerprint at spawn time, because none exists for it in that context; it is approved independently elsewhere. This is documented in `docs/AUTH.md`. It is exploitable only by an actor with write access to the host's data directory, which is outside this component's threat model. The complete fix is tracked in INTR-443 and is required before this path executes untrusted agent steps.

## Testing

The full test suite passes: 7,526 tests, no failures. This includes the end-to-end test that installs a package from a test registry, approves it, and runs it to completion, and a new test confirming that an onTrigger section runs its AI step and is rejected when its model configuration is missing. The UI end-to-end test also passes. The full branch was reviewed and approved.


alexanderguy:
## Branch status

This branch has been rebased onto current main, extended, and reviewed in full.

**Status: ready to merge.** The pull request is mergeable (it was previously in conflict). The full test suite passes (7,526 tests, no failures), along with the UI end-to-end test and a new end-to-end test for the change below. Each commit builds independently.

**Changes since the prior review**
- Rebased onto main's revised run-tracking model, in which a deployment is represented as its own run record. The code-sourced deployment's run-record write was reimplemented to match, including a guard that halts if the run's identifier and address are inconsistent. One database migration is added.
- Code-sourced onTrigger sections now run their AI step end to end. Previously they deployed but failed at runtime because their model was not provisioned. Each section's model is now provisioned from the approved definition using the same approval check as the parent workflow; a section without an approved model is rejected.

**Security review**
The security-critical code was reviewed and approved. The review confirmed that the change's reuse of an existing internal field does not alter unrelated behavior. Approval is bound to a fingerprint of the approved definition, and the code is re-verified against that fingerprint on each load.

**Tracked follow-ups (not blocking)**
- INTR-443: a workflow that references a separate, independently approved workflow is not re-verified against a fingerprint at spawn time, as none exists in that context. Documented in `docs/AUTH.md`; the complete fix is scheduled before this path executes untrusted agent steps.
- INTR-444: a cleanup to how a deployment record is rebuilt on configuration rotation.
- Credential-bearing model configuration is resolved per run rather than frozen; this is intentional.

PR #127, opened 2026-08-07
Extract the model catalog, add discovery probe automation, and onboard 26 models

alexanderguy:
## Summary

- Extract the hand-authored model catalog into `@intx/inference-catalog`, a standalone, zero-runtime-dependency, publicly-published package the dev seed consumes (layering dev-only pricing on top). Capabilities are baked as resolved literals with preserved provenance (`discoverySource` + `curatedCapabilities`); guard tests keep them derived from the discovery matrix and keep every offering's model present in the catalog.
- Add discovery probe automation so a new model can be onboarded without hand-rolling a script: `bin/probe` sweeps a model that need not be in the support matrix across every capability its provider can build, recording the HTTP outcome only the live call can see; `bin/classify-sessions` classifies the captures offline by replaying them through the production decoder. Supporting changes: a typed `CapabilityNotBuildableError`; google's model sets become a request-shape classifier rather than a discovery gate (an unknown chat model probes without pre-registration); `runCapture` returns the HTTP outcome and stops on a non-2xx; a replay-based semantic classifier maps a capture to the support-matrix outcome vocabulary.
- Onboard 26 net-new models across three providers via live probing plus replay verification (decode each capture through the production runtime, not a trusted HTTP 200): Anthropic (5), Gemini (2, dropping the retired gemini-3-pro-preview), and OpenAI (19, each with a per-model capability set — e.g. gpt-4 is text-only, gpt-4-turbo lacks structured output, and reasoning is unsupported on the first-party Chat Completions wire).
- Redact the OpenAI `openai-project` response header in captured fixtures, alongside the already-redacted `openai-organization`.

## Verification

- `make all` passes at HEAD: lint, `tsc` typecheck, admin-ui build, and the full test suite (0 failures).
- Every net-new model's capabilities are verified by replaying its captures through the production `runInference` decoder, so a baked capability means the runtime actually decodes real content — not merely that a request returned 200.
- A branch-wide fixture secret scan is clean: no real API keys, project or organization identifiers, or unredacted auth headers across any provider's sessions.

Closes INTR-445
Closes INTR-232


alexanderguy:

alexanderguy:
Branch extended since the initial review: two commits add optional per-offering seed pricing (priced XOR explicitly-unpriced) and the models already probed into the discovery matrix but not yet in the catalog (added as unpriced offerings on the existing providers). Whole-branch self-review (`code-review` skill) over the full range re-run and returned clean.

alexanderguy:
## Self-review

Whole-branch review (`code-review` skill) over `origin/main..HEAD` is clean. `make all` passes at HEAD, and no fixture across any provider carries a real API key or an unredacted project/organization identifier.

One known exception, tracked separately as **INTR-449**: the anthropic adaptive-thinking `output_config.effort` value differs between the runtime adapter (`high`) and the discovery rig (`max`), a pre-existing discrepancy out of scope for this branch.

PR #128, opened 2026-08-10
Reconcile the Anthropic adaptive-thinking effort as a deliberate pair

alexanderguy:
## Summary

- Names the runtime adapter's adaptive-thinking effort as
  `ADAPTIVE_THINKING_EFFORT = "high"` (the Anthropic API default) and the
  discovery capture rig's as `ADAPTIVE_THINKING_CAPTURE_EFFORT = "max"`, and
  documents that the two are a deliberate, unequal pair: production runs the
  API default, while the capture rig forces `max` because only `max` reliably
  elicits a capturable thinking block.
- Adds a guard test co-locating both values with that rationale, so a change
  to either surfaces the pair as a conscious decision.
- Raises the Node heap ceiling for the eslint lint step so the pre-commit
  hook's cold-cache lint does not run out of memory.

## Verification

- `make all` passes: lint, forced `tsc -b`, the admin-ui bundle, and the
  two-pass test suite (0 test failures).
- The new effort-pair guard test passes; the existing wire tests still assert
  the built request emits `effort: "high"` (runtime) and `effort: "max"`
  (discovery).

Closes INTR-449

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ build-admin-ui: FORCE
 
 lint: FORCE
 	prettier -c .
-	eslint --cache .
+	NODE_OPTIONS=--max-old-space-size=8192 eslint --cache .
 	$(BUN) bin/gen-api-docs.ts --check
 	$(BUN) bin/check-deps.ts
 	$(BUN) bin/check-launchers.ts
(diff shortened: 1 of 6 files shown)


alexanderguy:

The branch carries two intentional exceptions, each authorized by Greybeard:

- `packages/inference/src/providers/anthropic.ts` (`ADAPTIVE_THINKING_EFFORT`):
  the runtime effort constant is a bare untyped literal rather than typed
  against the effort union. Authorized by Greybeard on grounds that the entire
  runtime request body is untyped by design — every value is a bare literal
  validated by the live API and fixture replay — so typing one field would be
  an anomaly, not a fix; the value is pinned by the effort-pair guard test.
- The high/max rationale appears at four sites (the two constant comments,
  `docs/discovery.md`, and the guard-test comment). Authorized by Greybeard on
  grounds that the guard test is the anti-drift mechanism, each prose copy
  serves a distinct reader surface, and consolidating would require a fragile
  cross-reference the style skill discourages; it matches the model-set
  precedent.

PR #131, opened 2026-08-12
INTR - 252 - Make the iso git storage layer runtime agnostic

FarinM:
# Make iso-git storage runtime agnostic

## Why

`@intx/storage-isogit` was coupled directly to Node's `fs` and `path` modules. That prevented us from reusing the same Git-backed context, audit, mail, pack, and garbage-collection behavior in browser environments.

This change separates the storage implementation from its host filesystem. The package can now run against an injected runtime while preserving the existing Node and Bun behavior through a dedicated Node entry point. This also gives us a browser implementation backed by LightningFS and IndexedDB, which is the foundation for running agents with persistent local storage in the browser.

## What this adds

- A runtime-neutral `createIsogitStorage(runtime)` API. The runtime supplies isomorphic-git's filesystem client, path operations, atomic rename, and an optional persistence flush.
- Dedicated `@intx/storage-isogit/node` and `@intx/storage-isogit/browser` entry points. Existing server-side consumers now use the Node entry, while browser consumers can create an isolated IndexedDB-backed store by name.
- Normalized filesystem operations so the storage code works with both promise-based Node filesystems and the filesystem shape expected by isomorphic-git.
- Explicit durability boundaries around commits, branch changes, received packs, and garbage collection. Browser operations flush the LightningFS directory namespace before publishing refs or deleting superseded data.
- Crash-safe GC quarantine. Old packs and loose objects are moved into an unscanned `gc-trash` area before the replacement namespace is durably committed. They are deleted only after that namespace has been flushed to IndexedDB, ensuring that a reload cannot leave a ref pointing to missing objects.
- Browser integration coverage using real IndexedDB. The tests reload the page and interrupt persistence at key boundaries to verify that commits, refs, packs, audit history, and garbage collection remain readable.

## Result

The Git storage layer now has one shared implementation across Node, Bun, and the browser. Runtime-specific behavior is isolated at the package boundary, and browser persistence remains recoverable across reloads and interrupted writes while Node consumers retain the existing storage behavior through the explicit Node entry point.


packages/storage-isogit/src/node-metrics.ts
```diff
@@ -0,0 +1,80 @@
...
+  let fanoutDirs: string[];
+  try {
+    fanoutDirs = fs.readdirSync(objectsDir);
+  } catch (cause) {
+    if (hasCode(cause) && cause.code === "ENOENT") return 0;
+    throw cause;
+  }
+  let total = 0;
+  for (const name of fanoutDirs) {
+    if (!/^[0-9a-f]{2}$/.test(name)) continue;
```
alexanderguy, line 31: We've now got two copies of the repo-metrics helpers — the async runtime-based set in `repo-disk.ts` and the sync `node:fs` set here for the `dispatch-attribution` hot path. The sync need justifies the second copy, so no objection. One tidy-up: the loose-object filters aren't quite the same shape (`repo-disk` skips `pack`/`info` by name *then* applies the `/^[0-9a-f]{2}$/` regex; this one uses the regex alone). They produce identical counts — every name the explicit check drops already fails the regex — so it's harmless redundancy rather than a real gap. Might be worth a comment pointing each at the other so they stay in step.

packages/storage-isogit/src/node.ts
```diff
@@ -0,0 +1,80 @@
...
+  type MailCommitOptions,
+  type MailCommitResult,
+  type MailDirection,
+  type MailEntry,
+  type IsogitPath,
+  type IsogitRuntime,
+} from "./index";
+
+export const runtime = createNodeIsogitRuntime();
+const storageRuntime = normalizeRuntime(runtime);
```
alexanderguy: Small thing: `storageRuntime = normalizeRuntime(runtime)` and `createIsogitStorage(runtime)` each normalize, so the class and the free-function API wrap two separate `StorageRuntime` instances over the same fs. Fine as-is — the runtime is stateless and the lock is keyed by resolved path, not by instance. Only reason I raise it: the day someone adds per-runtime state, those two paths diverge silently. Same shape shows up one line down, where the store gets the unresolved `dir` while the `createIsogitStorage` path resolves it first. If `createIsogitStorage` handed back its normalized runtime, `node.ts` could bind a single one and sidestep both. Not worth churning the PR over.

packages/storage-isogit/src/runtime.ts
```diff
@@ -0,0 +1,277 @@
...
+  };
+
+  return {
+    git: runtime.fs,
+    access: async (filepath) => {
+      await lstat(filepath);
+    },
+    readFile,
+    readTextFile: async (filepath) =>
+      new TextDecoder().decode(await readFile(filepath)),
```
alexanderguy: Micro-nit, take it or leave it: a handful of spots (this one, plus `commitRecordFiles` / `readAt` / `readManifestHistory` in `store.ts` and `applyPack` in `pack-receive.ts`) each `new TextDecoder()` per call. A shared module-level instance would cover all of them.

packages/storage-isogit/src/store.ts
```diff
@@ -233,12 +249,18 @@ function decodeJsonlLines(text: string): unknown[] {
...
 async function readBlobAtCommit(
+  runtime: StorageRuntime,
   dir: string,
   oid: string,
   filepath: string,
 ): Promise<Uint8Array | null> {
   try {
-    const { blob } = await git.readBlob({ fs, dir, oid, filepath });
+    const { blob } = await git.readBlob({
+      fs: runtime.fs.git,
```
alexanderguy, line 259: Noticed `readBlobAtCommit`'s catch returns `null` for *any* error, not just absence — so a real I/O or corruption error during a time-travel read would read as "empty history." I know this predates the PR (you've only swapped `fs` → `runtime.fs.git` here), and I like that the new durable path got the strict variant. Flagging only so the tolerant behavior stays a deliberate choice rather than an accident. Fine to leave for a follow-up if you agree it's worth tightening.

alexanderguy (changes requested):
Reviewed by an agent via `pull-request-review`.

Read through commit-by-commit — structure and test coverage made this easy to follow. I ran `make build` (clean) and the storage suite (178 pass), so nothing here is about correctness.

Marking as requesting changes on the maintainability notes below — mostly the duplicated repo-metrics helpers and the double-normalized runtime in `node.ts`, since both are the kind of thing that drifts quietly later. Happy to see them addressed or waved off with a reason; either resolves my review.

alexanderguy (approved):
LGTM

PR #133, opened 2026-08-16
Fold child-workflow grant surfaces into the parent deploy approval

alexanderguy:
## Summary

- Store an approved grant surface per workflow definition version: the
  deployment-flattened union of a definition's own runtime-enforced
  (`tool:`/`effect:`) grants and its direct children's pinned surfaces
  (own ∪ ⋃ pinned(directChild)), so an operator approves the whole
  surface a run can reach through its children.
- At deploy, resolve each direct `childWorkflow` reference to its approved
  surface and fold it into the parent's pinned surface before the
  deployment becomes trigger-routable; a self-referential, wrong-kind,
  missing, or unapproved child aborts the deploy.
- Materialize a run's runtime ceiling from the stamped surface, read fresh
  per run at the anchor's version; a `childWorkflow`-bearing definition
  with no stamped surface fails closed at materialization instead of
  running unbounded.

## Verification

- `make all` passes: 780 tests pass, 0 fail across 123 files; lint,
  dependency check, and generated-doc check green; exits 0.
- The child-workflow fold, the fail-closed guard, and the per-run ceiling
  read are covered by DB-backed tests (fold resolution, frozen-ceiling
  staleness regression, and the approved-surface write-read seam).

Closes INTR-421

### DIFF
--- a/packages/db/migrations/0082_add_workflow_definition_version_approved_grant_surface.sql
+++ b/packages/db/migrations/0082_add_workflow_definition_version_approved_grant_surface.sql
@@ -1,0 +1,1 @@
+ALTER TABLE "workflow_definition_version" ADD COLUMN "approved_grant_surface" jsonb;
--- a/packages/db/migrations/meta/0082_snapshot.json
+++ b/packages/db/migrations/meta/0082_snapshot.json
@@ -1,0 +1,4102 @@
+{
+  "id": "3376fa93-1b95-4d59-9736-2556ee62eca6",
+  "prevId": "83d02430-cec6-45a5-a75b-dbaf491fa4bc",
+  "version": "7",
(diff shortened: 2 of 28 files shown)


alexanderguy:

Finding 1 from the whole-branch review was determined (Greybeard-confirmed)
to be separate work — a distinct gap that stands on its own — and is
tracked as INTR-458 (blocks INTR-419), not carried on this branch.

alexanderguy:
Closing this — it was building the grant fold against the native (by-id lookup) model, which is a dead end. INTR-421 is now blocked on retiring the native path (INTR-419); once childWorkflow is a real import, the fold falls out for free. See INTR-421 for the write-up.

PR #134, opened 2026-08-18
Source workflow definitions from hub git and package-registry assets

alexanderguy:
## Summary

- Sources a workflow definition's code from a hub asset — a package-registry tarball, a single git package, or a bun/npm workspaces monorepo — materialized by native git-subtree checkout at a pinned commit (git tree-oid content identity) indexed into a durable per-asset `.git`; deploy and boot-restore re-derive the closure from that store with nothing re-delivered.
- Keys workflow definitions on `(assetId, wireHash)` with no `packageName` column, so one asset backs many definitions: identical-surface members collapse to one row while each deployment still materializes its own source.
- Resolves a monorepo's workspace members and the `workspace:` / `catalog:` dependency protocols, producing a mixed closure of git-source members and external registry dependencies.
- Bounds an untrusted pushed pack's cost (object count, per-object inflation before `git.indexPack` allocates, cumulative checkout bytes/inodes), rejects a symlink or submodule tree entry at both the pack-receive push boundary and checkout, publishes materialized asset stores atomically (temp + rename), and classifies a structural asset reject as `path_violation` distinct from `corrupt`.

## Verification

- The build passes (`make` exits 0): lint, typecheck, and the full suite (7665 tests) are green.
- Source workflows install, deploy by source-ref, run to completion, and restore to routable from the durable git store alone — proven end to end through a real subprocess sidecar for a single git package, a monorepo member with a workspace dependency, a `catalog:` dependency, and a mixed workspace-local + external-registry closure.
- A tampered source deploy is rejected and never becomes routable; a credential-bound source workflow resolves and delivers its binding on deploy.
- Symlink and submodule tree entries hard-fail materialization with `pack.reject reason=path_violation` (the detail carries the offending path), for every asset kind — including the pre-existing `skill`-kind attach path, where a skill asset whose tree carries a symlink is rejected rather than materialized as a wrong-content regular file.

Closes INTR-417
Closes INTR-418
Closes INTR-420


alexanderguy:

The whole-branch review is clean. Two behaviors are deliberate and worth a reviewer's attention:

- **Symlink and submodule tree entries hard-fail materialization, for every asset kind.** The pack-receive boundary rejects them at push and `writeTreeToDisk` rejects them at checkout, both with `path_violation` (the reject detail carries the offending path). This also governs the `skill`-kind attach path: a skill asset whose tree carries a symlink is rejected rather than materialized as a wrong-content regular file.
- **The tool-package manifest source shape carries no DB migration.** `ToolPackageSource` nests an entry's bytes under a `package` envelope (a `tarball` or `source` arm); no durable row persists the prior shape across a reset, so no stored manifest depends on it.

A `pnpm-workspace.yaml` layout is rejected with a clear message; full pnpm support is tracked in INTR-461.

PR #136, opened 2026-08-19
Cover a single-source quota case, stabilize a flaky roundtrip test, and reconcile stale execution-host docs

alexanderguy:
## Summary

- The reactor suite asserts that a single source returning `quota_exhausted`
  is attempted once and then surfaces the error, locking in the
  wrapper-owns-retry contract that had only transitive coverage.
- The child-workflow roundtrip tests carry an explicit 180s per-test budget,
  so the fan-out case no longer times out on Bun's 5s default when the file
  runs standalone.
- The execution-host design docs match the shipped runtime: stale references
  to the deleted trivial/in-process symbols read in past tense, completed
  phases are marked landed, the isolation direction is recorded as
  capability-based placement (with exclusive sidecar allocation as the shipped
  substrate), and the deploy-frame field list is complete.
- The sidecar entry point drops a dead workflow-supervisor re-export, and the
  frame docs describe the supervised workflow child rather than the retired
  in-process harness.

## Verification

- `make all` passes on HEAD (lint, `tsc -b`, admin-ui bundle, 772 tests, 0 fail).
- The child-workflow roundtrip file passes standalone under Bun's default with
  no `--timeout` override — the exact condition that previously failed.

Closes INTR-385
Closes INTR-215
Closes INTR-264

### DIFF
--- a/apps/sidecar/src/index.ts
+++ b/apps/sidecar/src/index.ts
@@ -21,8 +21,6 @@ import {
   readRegistryMaxTarballBytes,
 } from "./config";
 import { createDefaultHarnessBuilder } from "./default-harness";
-// Pull the workflow-host wiring factory into the sidecar's module
-// graph so the supervisor surface is reachable from this binary.
 // `createSidecarDeployRouter` is the production routing the
 // orchestrator hands to the link's `agent.deploy` handler; every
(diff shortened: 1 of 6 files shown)


alexanderguy:

PR #137, opened 2026-08-19
Retire the static workflow.json definition path

alexanderguy:
## Summary

- Every workflow deploys from a code source — an npm registry package, an asset tarball, or an asset source tree. The hub probes the source in a sandboxed sidecar child, freezes an approved wire hash, and deploys by source ref; the workflow-process child re-evaluates the pinned closure for its runnable definition and re-verifies it against the hub-approved hash, failing closed on any divergence.
- The static `workflow.json` path is gone end to end: the live-authored deploy chain, the single-agent instance path, the `workflow.json` envelope push, and the sidecar's on-disk `workflow.json` are all deleted. A deployment's pinned definition is the frozen source closure, and run grants derive from a persisted grant-walk snapshot.
- Tool support runs on the code-sourced path (external `@intx/tools-*` packages, posix/LSP plugins, and credential-consuming tools), and the admin UI deploy form has a source picker for the three source variants.

## Verification

- `make all` passes: lint, the type-check over the whole graph, the admin-ui production bundle, and the full test suite.
- The code-sourced deploy works end to end against the real hub and sidecar — the source, tamper-reject, crash-restart, reconnect, and restore suites are green, and tamper-evidence holds (a closure that no longer projects to the approved hash fails closed). The frozen approval bundle routes intact through the router to an allocated worker on the exclusive-placement path, and a `childWorkflow` nested in an `onTrigger` body fails closed at runtime with the not-implemented error rather than a fabricated step success.
- The admin-ui source picker carries unit tests over the per-variant deploy request and Playwright specs through the deploy form for all three source variants (registry, asset tarball, asset source tree). A sidecar-backed admin-ui e2e drives the picker to a real, un-stubbed deploy and confirms the deployment reaches the deployed state on the hub.

Closes INTR-419


alexanderguy:
Self-review (`code-review` skill) returned clean over the full branch, including the test-coverage follow-on commits.

PR #138, opened 2026-08-19
Remove orphaned per-credential grants and reject in-use deletes

alexanderguy:
## Summary

- Credential deletion removes the credential and its per-credential
  `credential:{id}` grants in one transaction, so grants no longer outlive
  the credential they authorize.
- Deleting a credential or wallet still referenced by a model provider
  returns 409 Conflict instead of a raw 500 (the `model_provider` foreign
  keys are `onDelete: restrict`); a foreign-tenant id returns 404.
- A one-time migration removes `credential:{id}` grants already orphaned by
  pre-fix deletes, sparing the coarse `credential:*` role grant.
- `buildCredentialDelivery` no longer builds the unused `bindingGrants`.

## Verification

The build passes (`make all` exits 0). Credential and wallet deletion, the
409 in-use paths, tenant-scoped 404s (including a cross-tenant probe), and
the cleanup migration are covered by tests against a real migrated Postgres
schema.

Closes INTR-411

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -879,6 +879,7 @@ Only credentials owned by this tenant can be revoked.
 
 204: (no content) -- Credential revoked
 404: ErrorResponse -- Credential not found
+409: ErrorResponse -- Credential is in use by a model provider
 
 ## Discovery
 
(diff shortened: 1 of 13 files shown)


alexanderguy:

PR #139, opened 2026-08-20
Auto-respawn the workflow-host child with a crash-loop guard

alexanderguy:
## Summary

- The workflow-host supervisor watches `handle.exited` and auto-respawns the workflow-process child on an unexpected exit (crash, OOM, panic, SIGKILL). It replays mail stranded mid-flight back to the inbox tail first, so FIFO ordering holds across the respawn boundary. A clean child death does not trip the IPC crash callback, so the exit-watcher is the load-bearing signal.
- A crash-loop guard bounds respawns: 3 unexpected exits within 60s latch the deployment to a terminal `crash-looping` phase instead of respawning. On latch the supervisor commits a `RunFailed` for the deployment's run, so the run's externally-queryable status flips to `failed` — the durable surface, since `crash-looping` is in-memory only.
- Each respawn waits behind exponential backoff from 1s doubling to a 30s cap. A 60s stable run resets both the crash counter and the backoff.
- A shared respawn driver runs the kill/replay/respawn/install sequence for both the crash path and `recycle()`, so the two paths cannot drift.

## Verification

- `make all` passes: lint, type-check (`tsc -b`), the admin-ui production bundle, and tests (798 pass, 0 fail).
- Two integration tests run against a real sidecar subprocess with a real workflow-process child: one SIGKILLs the child mid-run and asserts the same sidecar respawns it, replays the stranded mail into FIFO position 0, and drives all three mails to completion in order; the other SIGKILLs the child three times and asserts the deployment latches to `crash-looping`, commits the observed `RunFailed`, keeps the sidecar alive, and spawns no fourth child.
- Unit tests cover crash counting and window pruning, backoff doubling and the cap, the stable-run reset with an injected clock, the generation guard against respawning a recycle-installed cohort, the supervisor-authored `RunFailed` commit, and push validation of a lone seq-1 `RunFailed`.
- The terminal-event to `workflow_run.status` DB flip is covered end-to-end against a real test DB, including the exact lone seq-1 `RunFailed` shape the crash-loop guard emits (`tests/db/workflow-run-terminal-flip.test.ts`).

Closes INTR-193
Closes INTR-469


alexanderguy:

A whole-branch review (`origin/main..HEAD`) found no correctness issue. The branch carries two intentional exceptions, each authorized by Greybeard:

- `apps/sidecar/src/workflow-host-wiring.ts` (the `activeSupervisors` map is pruned only on deploy-unwind and undeploy) and `packages/workflow-host/src/supervisor/supervisor.ts` (the crash-loop latch self-terminates with no host-facing signal): a supervisor that reaches a terminal phase on its own — the new `crash-looping` latch, or the pre-existing `stopped`-on-unexpected-exit path — stays in the sidecar's `activeSupervisors` map, so redeploying the same address needs a prior undeploy. Greybeard: this is a pre-existing lifecycle gap that predates and is broader than this branch (the old `stopped` path behaved identically), and the branch makes it rarer, not more common (a single exit now respawns and recovers). The acceptance-required observability surface — the durable `RunFailed` status flip — is delivered. Tracked for follow-up as INTR-470.

- `packages/workflow-host/src/supervisor/supervisor.ts` (`installNewChild` on the recycle path arms no stable-run reset timer, and a recycle touches neither `crashTimestamps` nor `respawnBackoffMs`): after crashes accumulate, an operator/policy recycle does not reset the crash counter or backoff; only the crash-respawn path and the 60s window pruning clear crash state. Greybeard: correct by design — a recycle is a planned replacement, not evidence the child is stable, so it must not launder recent crash history; the residual (an elevated backoff carried across a recycle until the next crash-respawn's own reset) is conservative in the safe direction. No defect, no follow-up.

PR #148, opened 2026-08-25
Fix and harden the deferred unified-host follow-up cleanups

alexanderguy:
## Summary

- The supervisor crash log interpolates the crash reason on both crash
  branches instead of emitting a literal `{reason}` placeholder.
- The admin-ui API client handles 202 Accepted responses: an empty-body
  202 resolves to `undefined` and a body-bearing 202 is parsed, rather
  than throwing "Unexpected end of JSON input".
- The cold-path approval-resume assertion requires a re-dispatchable
  (`suspendedCall`-bearing) approval gate for the resumed correlationId,
  mirroring the reactor's discriminator, so a mis-keyed resume fails
  loudly instead of silently skipping the approved call.
- The sidecar computes the per-step deploy directory through the canonical
  `agentDir` helper, and the dependency-light `@intx/hub-agent/paths`
  entry drops its now-unused `sanitizeAddress` re-export.
- Step teardown surfaces plugin and tool-bundle disposer failures as an
  `AggregateError` instead of swallowing a leaked LSP subprocess to a
  warning, and the two teardown callers route their close through a shared
  `runBodyThenCleanup` helper so a rejecting close never masks an
  in-flight step or control-loop error.
- Warm-agent eviction and source rotation act on every entry and
  aggregate failures, so one agent's failure never strands the rest.

## Verification

`make all` exits 0 (lint, type-check, admin-ui bundle, and both test
passes, run against a provisioned local database). Every change ships
with tests; the crash-reason, 202-empty-body, approval-resume keying,
disposer-failure masking, and warm-agent stranding paths are covered by
mutation-verified regression tests.

Closes INTR-216

### DIFF
--- a/apps/admin-ui/src/lib/api.test.ts
+++ b/apps/admin-ui/src/lib/api.test.ts
@@ -1,0 +1,61 @@
+import { describe, test, expect } from "bun:test";
+
+import { api } from "./api";
+
+// The hub's async-accept routes return 202. A signal route returns 202 with
+// an empty body; a trigger route returns 202 with a JSON acknowledgement.
+// api() must resolve both without throwing "Unexpected end of JSON input",
+// which res.json() raises on an empty body. Driving a real Response through
+// a stubbed global fetch exercises the actual body-reading branch rather
+// than a mock that pre-decides the result.
+describe("api 202 handling", () => {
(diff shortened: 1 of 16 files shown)


alexanderguy:

PR #149, opened 2026-08-25
Break runaway agent loops that repeat a tool call

alexanderguy:
## Summary

- The reactor detects when an agent repeats an identical tool-call turn N
  times in a row (N configurable, default 3) and ends the run with a fatal
  `reactor.error`, closing it as `message.run.ended` with `kind: "doom_loop"`.
  This guards against a runaway model that repeats an identical call: it makes
  no progress but keeps paying for an inference every turn.
- A turn's identity is the order-independent multiset of its executed calls'
  `(name, arguments)` pairs, canonicalized so reordered object keys and
  reordered parallel batches still match, while genuine retries with different
  arguments do not. Only calls that actually run count, so a suspended call that
  is approved and re-dispatched counts as the single execution it is.
- The threshold is operator-configurable through `createAgent` and reaches the
  reactor, which stays the single owner of the default and its validation.
  Passing `false` as the threshold disables detection entirely; a non-positive
  or non-integer value still throws, so only an explicit `false` disarms it.

## Verification

- `make all` passes: lint, type-check, the admin-ui bundle, and both test
  passes (unit and DB-backed integration) exit clean.
- The doom-loop suite exercises detection at threshold, no false positive on
  differing-argument retries, key-reordered equivalence, identical calls that
  error or succeed every turn, the suspend/approval/redispatch regression,
  repeated parallel fan-out, a single in-turn batch, counter reset on a
  different batch, threshold validation, and the disabled path. A seam test
  confirms the configured threshold reaches the reactor through the assembly.

Closes INTR-53

### DIFF
--- a/docs/INFERENCE.md
+++ b/docs/INFERENCE.md
@@ -407,7 +407,7 @@ It returns one or more actions. Multiple actions execute concurrently where poss
 
 **If the director throws an exception**, the reactor catches it, emits `reactor.error` with the exception details, and initiates graceful shutdown. The director is user-provided code and must not be able to crash the reactor without a clean terminal event.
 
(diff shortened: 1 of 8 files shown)


alexanderguy:
Self-review (`code-review` skill) of the full branch returned clean.

PR #150, opened 2026-08-25
Recover interrupted run-event compactions at supervisor boot

alexanderguy:
## Summary

- A single boot scan of `runs/` yields both the orphan replay's owned
  message ids and the pending-seal run ids, so recovery adds no second
  walk of the run tree.
- A serial, best-effort sweep re-seals runs left terminal-but-per-event
  when a crash interrupts the fire-and-forget compaction fold, reusing the
  idempotent `compactRunEvents` and reporting `{ sealed, failed }`.
- The supervisor runs the sweep at spawn off the dispatch path and awaits
  its settlement at shutdown; the compaction fold and the boot scan share
  one terminal-event authority so they cannot disagree on which runs are
  terminal.

## Verification

- `make all` passes: lint, `tsc -b --noEmit --force`, the admin-ui bundle,
  and tests (7026 unit + 820 integration, 0 failures).
- New unit tests exercise the crash window end-to-end on a real temp-dir
  substrate: a terminal-but-per-event run seals to a byte-exact combined
  file, live and already-sealed runs are no-ops, a sweep racing a live fold
  seals exactly once, and a failed fold is reported without aborting the
  sweep.

Closes INTR-229

### DIFF
--- a/packages/hub-sessions/src/index.ts
+++ b/packages/hub-sessions/src/index.ts
@@ -156,6 +156,7 @@ export {
   dequeueToProcessing,
   readProcessingEntry,
   markConsumed,
+  scanRunsForBoot,
   readOwnedMessageIds,
   readCommittedWorkflowRunLifecycle,
   readWorkflowRunLifecycle,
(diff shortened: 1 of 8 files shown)


alexanderguy:

PR #151, opened 2026-08-26
Reclaim a self-terminated workflow-host supervisor

alexanderguy:
## Summary

- A workflow-host supervisor that self-terminates — the crash-loop guard
  latching to `crash-looping`, or an unexpected child exit or recycle failure
  reaching `stopped` — surfaces an `onSelfTerminate` binding that fires only on
  genuine self-termination, never on a host-requested `shutdown()` or an
  initial spawn failure.
- The sidecar deploy router subscribes to it and reclaims the deployment
  address: it drops the supervisor from `activeSupervisors` and releases the
  address's transport registration, routers, slug, and deployment mapping, so
  the address redeploys with no manual undeploy first.
- The reclaim leaves durable state (the run record, on-disk scratch) untouched,
  mirroring hub deployment intent, and is idempotent, so it cannot race a
  concurrent operator undeploy.

## Verification

- `make all` passes build, lint, type-check, and docs. The workflow-host
  supervisor and sidecar wiring suites pass. (The repository's Postgres-backed
  integration tests are unavailable in this environment for lack of a `DB_HOST`
  database; that is pre-existing and unrelated to this change.)
- Tests cover the signal firing once per self-termination path and staying
  silent on `shutdown()` and on a pre-ready spawn crash, and the sidecar
  reclaiming a self-terminated address so a redeploy succeeds without a prior
  undeploy, idempotently against a following undeploy.

Closes INTR-470

### DIFF
--- a/apps/sidecar/src/workflow-host-wiring.test.ts
+++ b/apps/sidecar/src/workflow-host-wiring.test.ts
@@ -31,6 +31,7 @@ import {
   validateWorkflowProjection,
 } from "./workflow-host-wiring";
 import {
+  createDeploymentAddressRegistry,
   createMultistepGrantsRouter,
   createMultistepMailRouter,
   createMultistepSourcesRouter,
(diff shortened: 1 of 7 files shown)


alexanderguy:

PR #152, opened 2026-08-27
Give the warm agent a conversational inbound-mail loop

alexanderguy:
## Summary

- Gives every workflow run an IMAP-esque mailbox in its own git-backed
  workflow-run substrate, replicated to the hub. A new `@intx/mailbox` package
  holds the IMAP model (UID, flags, search, thread, fetch), extracted from
  `@intx/mail-memory`; a substrate-backed store persists it. The warm
  single-step agent reads this mailbox locally through `mail_read`,
  `mail_search`, and `mail_wait` with no hub round-trip. Message bodies are read
  from disk on demand, so the resident footprint stays bounded to metadata.
- The supervisor eager-commits each inbound message into the run mailbox on
  arrival and notifies the child over the control channel, so a tool call in the
  same turn sees the new mail. Flushes persist incrementally through a substrate
  tree delta rather than rewriting the whole subtree.
- The warm agent auto-replies through the existing signed outbound bridge. A
  `connector.reply` composes a reply threaded by `In-Reply-To` and the full
  `References` ancestry, resolved from the mailbox. The reply is at-least-once
  durable: the inbound mail is settled only after the send is acked, so a crash
  before the ack replays the mail. `createHarness` and the warm path share one
  reply-drain.
- Both conversational shapes work: an interactive step (`triggers: "unbounded"`)
  where each inbound mail is a connector-seeded dispatched turn, and a batch step
  that holds one run open and blocks on `mail_wait`.
- The warm agent manages and consumes its mailbox through two new tools:
  `mail_flag` sets or clears IMAP flags (one direction per call, so an error
  means the mailbox is unchanged) and `mail_expunge` sweeps every `\Deleted`
  message, returning the uids it removed. Both route through the supervisor --
  the sole mailbox writer -- over a `mailbox.mutate` control frame, so the child
  never flushes the run ref and never races the supervisor's mirror. Expunge
  physically removes a message's `<uid>.eml` from the live tree; the bytes
  survive in git history because a `workflow-run` repo's objects are never GC'd,
  an invariant a tripwire test pins at the GC decision site (adding the kind to
  the GC allow-list would fail the build).

## Verification

- `make all` passes on the full branch (lint, `tsc -b --force`, admin-ui bundle,
  tests): 7126 fast tests and 839 slow tests green, 0 failures. Unit suites cover
  `@intx/mailbox` (disk-on-demand reads, search, thread, fetch), the substrate
  mailbox store (incremental delta flush, no resident bodies), the mailbox-notify
  IPC arm and watch registry, the reply-drain with full-`References` resolution
  and the durable send barrier, the mailbox-mutation frame + child bridge, the
  supervisor mutation handler (addFlags / removeFlags, single / multi / empty
  expunge sweeps, non-INBOX rejection), the transport reroute, the widened
  `expunge` return, the `mail_flag` / `mail_expunge` tools, and the relaxed
  replication check (a deletion is accepted, a retained blob's mutation still
  rejected, the index must persist, `workflow-run` stays off the GC allow-list).
- End-to-end against a real hub and a real sidecar subprocess (under
  `tests/workflow-deploy/`): the auto-reply's `In-Reply-To` equals the inbound
  `Message-Id` and its `References` carries the full ancestry on the wire;
  `mail_wait` wakes mid-turn on newly-arrived mail; a two-turn conversation
  threads both within one run via `mail_wait` and across dispatched turns for an
  unbounded step; and the warm agent flags a message `\Deleted` then expunges it
  -- a second, unflagged message survives the sweep (the expunge honors
  `\Deleted` selectively), and the expunged `<uid>.eml` is gone from the tip tree
  but reachable through an ancestor commit (audit bytes retained in history).
- Scoped out, by design: the hub read route `GET /:runId/mail` is future work
  (INTR-454).

Closes INTR-480


alexanderguy:
Self-review (`code-review` skill) returned clean across the full branch, including the setFlags/expunge follow-up; no waived exceptions. Scope detail is in the PR description.

PR #153, opened 2026-08-27
Run a real per-step agent for childWorkflow fan-out

alexanderguy:
## Summary

- A `childWorkflow` fan-out step runs a real, tool-bearing per-step agent through the sidecar source-tools invoker, replacing the fail-loud `ChildStepNotImplementedError` stub. Each child step resolves inference against per-child-step sources staged at parent deploy (`assets/workflow/<childRef>/sources.json`); the parent run's live inference-event sink and a monotone spawn-depth ceiling thread down the terminal spawn seam.
- A spawned child's inherited grants are capped to what the child body itself declares — an in-process re-walk of the child definition plus a pattern-coverage filter (keep every `deny`/`ask`, keep an `allow` only when its pattern covers a declared resource). The parent stays the ceiling; the capped set is applied uniformly across the child's steps and written to both grant sinks, so a grandchild's ceiling is the capped set rather than the raw parent set.
- The design-of-record (`docs/unified-execution-host-design.md`) and the childWorkflow test suites reflect real child execution.

## Acceptance coverage

- Roundtrips: single child, nested to depth 2, 5-way siblings, a tool-bearing child (tool exposed), a real tool **invocation** landing in a child turn, a `map` fan-out nested inside a child body, and a child's live inference event reaching the hub.
- Grant cap: unit tests for the pattern-coverage filter, plus an end-to-end roundtrip proving a declared grant survives and an undeclared one is dropped.
- Depth ceiling: a runLocal unit test, a filter unit test, and a real-path test that drives an accumulating nested chain through `createSidecarRunChild` until the guard fires and names the offending depth.

## Verification

- Build and lint are clean; `make all` passes — 7149 unit tests and 846 workflow-deploy tests, 0 failures.
- All childWorkflow suites pass against the real hub + sidecar subprocess + mock-inference harness (and the sidecar-substrate unit harness for the grant-cap and depth tests).
- A pre-existing, non-deterministic flake in the shared mock inference server (`tests/hub-agent/lib/deploy-flow-env.ts` — `req.json()` on an empty-body request) can intermittently fail the full concurrent run. It is unrelated to this branch: that file is untouched, and the affected test passes in isolation.

Closes INTR-310


alexanderguy:
Self-review (`code-review` skill, whole-branch `origin/main..HEAD`) returned clean — no correctness defect. Cross-commit seams verified: the spawn-depth counter threading, the per-child-step sources-ref keying shared with the onTrigger-body path, the grant cap's interaction with the credentials snapshot and the grandchild ceiling, and that the removed stub and flat grant inheritance left no un-re-established invariant.

Two non-blocking observations, each a theoretical-only case in an unreachable path (kept as-is):

- `apps/sidecar/src/workflow-substrate-factory.ts`: `writeChildRunGrants` runs before the `dataDir`/`onEvent` guards, so an orphan `grants.json` could be written if a guard fired — but both are always present in production and a retry re-writes idempotently.
- `apps/sidecar/src/workflow-substrate-factory.ts`: an undefined `closurePackageDir` would surface as a raw capability-walk error rather than a targeted one — but that path is unreachable on the only lineage that stages childWorkflow assets.

PR #154, opened 2026-08-28
Track the README version from the git tag and skip DB tests without a database

alexanderguy:
## Summary

- The README version badge reads the latest release tag from GitHub (a
  dynamic shields.io tag badge) instead of a hardcoded series, and the two
  prose mentions drop the version number while keeping "alpha" — so the git
  tag is the single source of truth and nothing drifts.
- Database-dependent `tests/workflow-deploy/` suites skip instead of failing
  on a checkout with no configured database: a file-scope `beforeAll` returns
  early when the DB env is absent, and the e2e suites that lacked a guard wrap
  their `describe` in `skipIf`. The `harnessDbEnvAvailable` docstring now
  documents both shapes and the file-scope-hook trap that caused the drift.

## Verification

- `make all` passes with a database present (exits 0: lint, type check,
  admin-ui bundle, and full test suite green).
- `make test` passes on a checkout with the database env removed (exits 0; the
  previously-failing workflow-deploy suites now skip, 0 fail).

Closes INTR-486

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![CI](https://github.com/faremeter/interchange/actions/workflows/ci.yml/badge.svg)](https://github.com/faremeter/interchange/actions/workflows/ci.yml)
 [![Status: alpha](https://img.shields.io/badge/status-alpha-orange.svg)](#project-status)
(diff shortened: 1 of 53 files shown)


alexanderguy:

PR #155, opened 2026-08-29
Let loop bodies suspend and resume like onTrigger sections

alexanderguy:
## Summary

- Route each loop iteration through the shared suspendable-child seam — the same mechanism onTrigger section bodies use — instead of the in-process `createLoopIteration` path, so a loop body can park on an `awaitSignal` and resume across a sidecar crash.
- Lift the loop body-ban for `awaitSignal` and `childWorkflow` (`loop`, `onTrigger`, and `sleep` stay banned), and thread a child-spawn depth ceiling through the seam so a `childWorkflow` grandchild inside a loop body is bounded.
- Materialize each loop iteration run's own capped grants file at birth, write-once, so a `childWorkflow` grandchild spawned from a loop body is authorized on the deployed path.
- Tear an in-process suspendable child (loop iteration, onTrigger section body, or terminal `childWorkflow`) down locally on a parent-cascade abort — it settles `RunFailed` under its own principal rather than self-writing a supervisor `CancelRequested` it cannot sign through the run proxy — so a drain landing on a parked loop container sheds cleanly instead of hanging. The teardown cause is recorded durably (`ChildCompleted.abortedTeardown`) so a `tolerate` onTrigger section torn down by a drain stays ended on crash-resume rather than resurrecting.

## Verification

- `make all` passes: lint, `tsc -b`, the admin-ui production bundle, and the full test suite including the deploy e2e suite.
- A deployed loop whose body parks on an `awaitSignal` resumes across a real sidecar crash and restart and runs to completion, with the pre-park step applied exactly once.
- A deployed loop body spawns a `childWorkflow` grandchild that runs a real agent and completes.
- A deployed drain landing on a parked loop iteration, and on an in-flight `childWorkflow` grandchild, sheds the run to `RunFailed` without hanging.

Closes INTR-478


alexanderguy:

alexanderguy:
Follow-up self-review (`code-review` skill): the branch now also carries a commit making an in-process suspendable child (loop iteration, onTrigger section body, terminal childWorkflow) tear down locally to `RunFailed` on a parent-cascade abort instead of self-writing a supervisor cancel it cannot sign through the run proxy — fixing a deployed drain-shed hang — with the teardown cause recorded durably (`ChildCompleted.abortedTeardown`) so a `tolerate` onTrigger section stays ended on crash-resume. A whole-branch review returned clean.

PR #156, opened 2026-08-30
Allow a loop body to contain a nested loop

alexanderguy:
## Summary

- A loop body may contain a nested loop, with deterministic per-iteration child identity: a body-child run id carries its container run id (`<runId>__<loopId>__<index>`), the body enumerator and the deploy-time source pin recurse through nested bodies, and a step id may not contain the `__` ref delimiter so every derived id stays injective. The definition-time prohibition is lifted; `sleep` and `onTrigger` stay banned in a loop body, and a definition-time limit bounds static nesting depth.
- A nested loop's signal park composes: the container relay surfaces its await to its own parent, so a park relays one layer at a time to the run's real channel and a delivery cascades back down; resume and cancel propagate through every nesting level.
- Fixes a pre-existing single-level resume deadlock: a loop body signal gate parked when the crash landed before the container's relay await flushed hung on resume; it now recovers the parked gate and re-drives the relay.

## Verification

- `make all` passes (lint, type-check, admin-ui bundle, both test passes; 0 fail).
- Regression tests cover nested-loop convergence, exhaustion, carry, crash-resume (fully-parked and both partial-park windows), drain/cancel propagation, and the loud refusal of the one unsupported topology, at the runtime level and as a deployed roundtrip.

Closes INTR-274


alexanderguy:

PR #157, opened 2026-08-30
Author the dispatch orchestrator on the workflow engine

alexanderguy:
## Summary

- The outer per-level iteration and the Phase-5 build/attribute/fix/rebuild
  verification loop run as engine-native nested loops. The Phase-5 fix phase is
  depth-3: a per-task fix loop, each task retried by its own nested retry loop,
  with build and rebuild git effects inside the loop body and a per-task
  operator escalation via `awaitSignal`.
- The seven host on-disk resume detectors are expressed through the engine's
  journal-based resume (with a case-to-mechanism mapping table), and
  exactly-once effects hold across a crash inside the Phase-5 loop body.
- The full composition is proven end-to-end by a stubbed-invoker routing/resume
  test and a real-agent test through the production step-invoker; each
  crash-resume claim is backed by a defeated-ledger non-vacuity probe.
- Three Postgres-backed loop deploy suites skip cleanly on a checkout with no
  database env. A file-scope `beforeAll` runs even when `describe.skipIf` skips
  the suite bodies, so each hook carries its own `harnessDbEnvAvailable` guard,
  matching the other DB-backed deploy suites.

## Verification

- Lint, `tsc -b --force`, the admin-ui bundle, and the full test suite pass;
  `make all` exits 0 on a checkout with no database env. The Postgres-backed
  integration suites skip when `DB_HOST` is absent and run where a dev DB is
  configured.
- The dispatch orchestrator is proven by the two new tests:
  `tests/workflow/dispatch-orchestrator.test.ts` (12 tests) and
  `tests/workflow-deploy/dispatch-orchestrator-real-agents.test.ts` (4 tests).

Closes INTR-275


alexanderguy:

alexanderguy:
Follow-up commit `49f08e82` restores the issue's "per-task fix loop with per-task retries -- rework nested inside rework" as genuine depth-3 nesting (`phase5 -> taskLoop -> retryLoop`) in both the shape test and the real-agent test, replacing the earlier collapsed single fix loop. The shape test witnesses the nesting iterating (the phase5 fixer runs 2 tasks x 2 retries), and the real-agent test drives the retry loop through a second attempt with the real fixer. Reviewed clean.

PR #158, opened 2026-08-31
Resume a run parked mid-sleep across a crash

alexanderguy:
## Summary

- A run parked on a bare `sleep` (reduced phase `awaiting-timer`) resumes
  across a crash or sidecar restart: `runSleep` re-adopts its unfired durable
  timer (or completes an already-fired one) rather than minting a second timer,
  and a new `isResumableSleepStep` carve-out admits the step both on the DAG and
  in the resume classifier.
- An onTrigger body parked mid-sleep resumes as well: the container's resume
  planner re-adopts an in-flight body that has no container park, and still
  refuses a body parked on an un-relayed signal (author or control-plane), which
  onTrigger cannot relay and would otherwise hang.
- The loop and onTrigger resume planners share a single parked-signal scan.

## Verification

- `make all` passes (lint, type-check, admin-ui bundle, and the unit suite
  exit 0).
- The onTrigger-body case works end-to-end against a real sidecar crash and
  restart (`on-trigger-body-sleep-restart-resumes.test.ts`): a body parked
  mid-sleep completes only after the restart, with a single `TimerSet`/
  `TimerFired` pair on the body run.

Closes INTR-485


alexanderguy:

PR #160, opened 2026-09-01
INTR-483 - Add capability based sidecar provisioner selection

FarinM:
## Why

The shared/exclusive placement model encoded one infrastructure topology in workflow and tenant configuration. It could not express the guarantees a workload actually needs, made additional provisioner types difficult to introduce, and gave an ambient sidecar a broader authentication and routing role than any one deployment required.

Workflow source also has to be evaluated before its runtime requirements are known. That probe should run on explicitly provisioned capacity under tenant and operator policy, then hand off to a deployment selected from the frozen requirements. Probe and deployment infrastructure need the same durable, generation-fenced lifecycle so a reconnect or uncertain provider operation cannot cross allocation boundaries.

Finally, deployment callers should select catalog offerings rather than submit resolved provider credentials. The Hub can then authorize and resolve the inference chain under the deployment authority, keep secrets out of probe and launch records, and observe credential rotation or revocation during recovery.

## What this adds

- **Capability-based provisioner selection.** Workflows express `require` and `block` rules while provisioners advertise capabilities as `available` or `blocked`. Exact, namespace-prefix, and global selectors use specificity rules; narrower declarations win and blocking wins at equal specificity. An omitted capability is unknown, so selection fails closed instead of assuming a guarantee.
- **Independent policy enforcement.** Every tenant ancestor is evaluated as its own constraint, preventing a descendant from weakening inherited policy. Requirements from inline loops, trigger bodies, and child workflows are folded into the frozen wire identity and included in final placement.
- **Provisioned workflow probes.** Deployment first creates a minimal `workflow_probe` and provisions it through the first registered backend matching tenant and Hub probe policy. Registration order is the operator-defined selection priority, while Hub-level probe rules can require provider-neutral guarantees such as workload isolation without becoming workflow runtime requirements.
- **Capacity adoption when safe.** After the probe result is frozen, the Hub selects the deployment provisioner from tenant and workflow requirements. It adopts the connected probe capacity when the same binding still matches; otherwise it destroys that capacity and creates a normal pending allocation through the selected backend. Provisioners continue to use the idempotent `ensure`/`destroy` lifecycle and remain responsible for isolation and backing-capacity reuse.
- **Allocation-scoped trust and routing.** The Hub no longer starts or trusts an ambient sidecar. A provisioner-generated bearer credential resolves to one probe or workflow allocation generation, and probe, deploy, reconnect, mail, signals, initialization packs, and workflow-run packs are fenced to that identity.
- **Durable cleanup and recovery.** Probe rows retain only probe input/output and the binding needed for cleanup. Startup moves interrupted probes to `releasing` and attempts to destroy their capacity, while periodic reconciliation retries failed cleanup. Frozen launch specs store catalog offering IDs rather than provider secrets and resolve them again at allocation launch or recovery. Provisioned dispatch and approval lifecycle reads use the canonical run ID derived from the deployment address.
- **Catalog-based deployment input.** The API, client, and admin UI now submit ordered catalog offering IDs. The Hub resolves and validates the chain under the source authority before preparing the deployment.
- **Injectable Hub composition.** Operator compositions can register provisioners and probe capability policy at the Hub boundary. The production composition intentionally registers no infrastructure backend and fails closed; the local-process provisioner in this change is test-only.
- **Removal of legacy placement state.** Migrations delete legacy shared sidecar identities, remove the obsolete credential scope, and drop the shared/exclusive placement columns.

## Verification

- `make all`


alexanderguy (commented):
Reviewed with the `pull-request-review` skill — this was an agent-driven review, run over a worktree of the branch, with findings independently re-verified by a second pass. Flagging that up front for transparency.

Design looks solid. The capability model fails closed, the probe lifecycle reuses the existing allocation reconciler instead of a parallel one, and `make all` is green here. I couldn't find a fail-open hole in selection, and I confirmed the per-deployment identity model doesn't cap a physical sidecar to one deployment — that stays a provisioner concern.

Grouping the notes below by file since a few point at symbols rather than single diff lines. Most of this is "write it down" or small tightening, not rework.

---

**`workflow-allocation-service.ts` — `ensure()` -> `markProbing`**

`externalRef` only lands at `markProbing`, after `ensure()` has already created capacity. If the Hub dies in that window, restart cleanup calls `destroy()` with `(allocationId, generation, sidecarId)` but no `externalRef`. The local-process provisioner is fine — it keys on `allocationId` — so nothing leaks today. But that's now an implicit requirement. Can we state in `IMPLEMENTATION.md` that `destroy()` must succeed from `(allocationId, generation, sidecarId)` alone, so the next provisioner author doesn't rely on `externalRef`?

**`sidecar-token-authenticator.ts` / `HARNESS_DESIGN.md` — reconnect factor**

Dropping the challenge/response makes the allocation token a bearer secret — holding it is enough to reclaim the route and its mail/event stream. Blast radius is one allocation, so I'm not worried about ambient trust, but it does lean harder on TLS in transport and never logging the token. Assuming that's intended? Just want it to be an explicit call rather than a side effect of the doc rewrite.

**`session-service.ts` / deleted `tests/db/anchor-before-frame.test.ts`**

The other two deleted DB tests are fine — their functions are gone. This one's different: the anchor-before-frame ordering it proved is still live in `emitSourceRefDeployFrame`, and `session-service.test.ts` still points at "the tests/db regression test" that no longer exists. Looks like it got dropped because the emit moved to `sendAgentDeployToAllocation`. Could we rewrite it against the new path instead of losing it? Without it, an emit-before-insert regression passes green and surfaces as `path_violation` in prod.

**`sidecar-handler.ts` — `validateSidecarIdentity` default**

Default is `async () => true`. Prod wires the real `isCurrent`, so we're covered, but a future composition that forgets to wire it silently turns off scope revalidation. Can we default to reject so the safe path is free?

**`sidecar-capabilities.ts` — `collectPrimitiveRules`, `{ ref }` arm**

The `{ ref }` skip is correct today because the projection runs on the inline authored definition before extraction, so a ref'd body is a separate deployment that collects its own capabilities. The other ref-skips in the projector have a comment saying why they're safe; this one doesn't. Could we add an assert (or comment) that fold-time bodies are inline? If a future path ever folds a ref'd body that runs co-located, it'd under-count silently.

**`workflow-allocation-service.ts` — `allocationFences`**

We write a fence per probe and never delete them, so this grows with lifetime deployments. I think that's deliberate — it's a monotonic guard and deleting reopens the stale-reconnect window — but probes make it grow faster than allocations did. Is unbounded growth acceptable, or do we want a tombstone that still rejects a backward move?

**`sidecar-capabilities.ts` — fold dedup**

Nested rules get concatenated without dedup, so a capability required by both a parent and a child appears twice in the projection. Doesn't change matching, but it makes the wire hash depend on nesting shape. Worth a dedup before we freeze it?

**`sidecar-capabilities.ts` (types) — `parseSidecarCapabilitySelector`**

`"runtime:"`, `"a::b"`, and `":*"` all parse clean (verified). Not fail-open, but they're not real capability names either and they get past the boundary. Tighten the narrow to reject empty segments?

**`workflow-allocation-service.test.ts` / `sidecar-allocation-store.ts` — adopt seam**

The adopt disconnect -> reconnect -> re-auth seam is the load-bearing part of adoption, but the only adopt test stubs the router out, so that path isn't exercised. Could we add one integration test at the `createSidecarRouter` seam? A direct `createAdopted` state assertion would be a bonus.

**`SIDECAR_PLACEMENT.md`**

Two small things: (1) the `browserOnly` example would trip up an operator — `block runtime:*` is only satisfied by a `runtime:* blocked` declaration, and the example primes exact declarations, which would never match. One line spelling that out helps. (2) The Selection steps are ordered opposite the code — the no-candidates check runs before the chooser, and tenant policy resolves once before probing, not after the freeze.

**`DEV.md` — `bin/db-reset` table**

The `--clean` row still says it "wipes the hub and sidecar on-disk state", but the prose nearby was updated to "provisioner-managed local sidecar data". Quick sync so they agree.

**`wire-workflow.ts` / `tenants.ts` — `sidecarPlacement` field**

The field's still called `sidecarPlacement` but it now holds a capability policy. Fine as an umbrella name, but worth a deliberate keep-or-rename rather than a carry-over — your call.

---

Optional, pre-existing (not touched by this branch — fair to skip):

**`sidecar-handler.ts` — `handleMailOutbound` / `agent.event`**

Not something this PR changed, but now that each sidecar is bound to one deployment: `mail.outbound` doesn't check `senderAddress` against the connection, and `agent.event` isn't ownership-gated the way `connector.state.changed` is. Might be a natural tightening to fold in while we're here — your scope call.

alexanderguy (commented):
Re-reviewed the update (agent-driven again, via the `pull-request-review` skill, with the findings independently re-verified). Everything actionable from the first pass is resolved — thanks for the quick turnaround.

Confirmed fixed: the destroy-by-identity contract note, the bearer-token/TLS docs, the fail-open `validateSidecarIdentity` default (now a required field), the fence retirement (checked hard — no stale-reconnect window opens; the post-await fence re-check plus the durable `isCurrent` backstop close it), the capability-rule dedup, the empty-segment selector rejection, the real-router adopt-seam test, the three doc fixes, and the outbound-frame ownership gating. The anchor-before-frame ordering is also covered again by a real-DB test on the provisioned path.

Two things left, neither blocking:

1. **Could we drop `AgentKeyStore.signChallenge` (and the rest of the challenge-signing machinery)?** This PR retires the challenge/response reconnect flow, which leaves `signChallenge` with no production callers — only its own definition and tests reference it now. Since removing that flow is the theme here, cleaning up the orphaned signer fits this change (or a same-author fast-follow if you'd rather keep the diff tight). Flagging it so it doesn't linger as dead code.

2. Separately: I noticed a pre-existing issue in the reconciler — `destroyCurrent` ignores a `retryable: false` destroy rejection and retries forever, unlike `acceptEnsure` which fails terminally. It's byte-identical on `main`, so it's not yours to fix here — I filed it as INTR-503 so it's tracked on its own.

The one optional nit I raised earlier (a hard assertion instead of a comment on the `{ ref }` fold skip) is fine to leave — the fold only ever runs on the inline authored definition, so it's a future-proofing preference, not a correctness gap. Consider it closed.

Looks good to merge from my side.

alexanderguy (commented):
One more cleanup in the same vein as the challenge-signer removal, thanks for catching that one.

Retiring the challenge/response reconnect flow left `lookupPublicKey` orphaned the same way `signChallenge` was. On `main` it had four callers in the sidecar handler (the reconnect ownership-gate); this branch removed all of them, so it's now defined and declared on `SidecarLookups` but never called. Could we drop it while we're here?

- Remove `lookupPublicKey` from `hub-session-lookups.ts` and its entry on the `SidecarLookups` type.
- The doc comment on it still describes the "reconnect ownership challenge [that] fires in the deploy->first-trigger window" — that flow is gone, so the comment now contradicts the reconnect docs. Same for a handful of test comments (the deploy-flow harness, the reconnect roundtrip tests, the signal-correlation test) that still narrate an ownership challenge answered by `lookupPublicKey`.

The code itself is fine at runtime — reconnect works entirely through identity revalidation + allocation fences now — so this is dead-code and doc coherence, not a behavior fix. Grouping it with the signer removal keeps the challenge-era leftovers from lingering.

alexanderguy (approved):
LGTM

PR #161, opened 2026-09-01
Cover a credential-bound workflow deploy through the HTTP route

alexanderguy:
## Summary

- An integration test deploys a credential-bound definition through the
  `POST /workflows/deployments` route on the shared path against a real
  session service, asserting the binding resolves and decrypts; a wrong-key
  cipher and a keyless (noop) app each fail the deploy closed (502).
- The noop credential cipher rejects a real `enc:` ciphertext on decrypt
  instead of passing it through, so a keyless composition fails closed
  rather than delivering an un-decrypted secret.
- The model-source resolution path takes the credential cipher as a required
  parameter resolved at the edge, rather than defaulting to a noop deep in
  the resolution code.

## Verification

- `make all` exits 0 (7222 + 873 tests, 0 fail).
- The integration test drives the real route against real Postgres and a
  live sidecar; the positive deploy returns 201 (binding resolved and
  decrypted), and the wrong-key and keyless deploys return 502. Removing the
  cipher forward from the route flips the positive deploy to a fail-closed
  502, so the test guards the route-to-resolution hop.

Closes INTR-499


alexanderguy:

alexanderguy:
Follow-up commit (require the credential cipher on the model-source resolution path): self-review (`code-review` skill) returned clean.

alexanderguy:
Added a keyless-app route deploy test (a hub with no CREDENTIAL_ENCRYPTION_KEY fails a binding-carrying deploy closed on an encrypted credential). Self-review returned clean.

PR #162, opened 2026-09-01
Support loops nested in onTrigger and childWorkflow bodies

alexanderguy:
## Summary

- A `loop` nested inside a spawned workflow body — an onTrigger section or a childWorkflow child — deploys and runs to completion. The deploy-time per-step inference-source pin recurses into the loop body to pin its agent step's source, and the spawned-body child runtime env wires loop support (`loopFns` + `spawnLoopIteration`).
- The per-step source pin is a shared `pinInertStepSources` walk that both pin sites drive with their own leaf policy; a body agent that declares no source is approval-gated rather than pinned to an unapproved placeholder.

## Verification

- `make all` exits 0, including the real-DB integration suite.
- A deployed e2e drives an onTrigger section whose body contains a loop with an agent body to completion, including a loop body that spawns a childWorkflow grandchild.

Closes INTR-477


alexanderguy:

PR #163, opened 2026-09-01
Add a concurrent-signal validator, share the run-lifecycle readers, extract a terminal-flip helper, and remove the dead deploy-test tarball wiring

alexanderguy:
## Summary

- Reject at authoring time any workflow whose two schedulable-concurrent
  `awaitSignal` steps share a signal name, spanning the loop/onTrigger relay
  seam (childWorkflow excluded); the runtime fail-loud guard stays the backstop.
- Consolidate the two run-lifecycle readers (working-tree fs and committed git
  objects) behind one shared classification core fed by per-surface read
  adapters; `scanRunsForBoot` stays separate.
- Extract the duplicated run-phase→terminal-status mapping into a pure
  `decideTerminalRunFlip` helper and a named `TerminalRunPhase` type.
- Remove the vestigial synthetic tools-mail and credential tarball builders from
  the deploy test harness, along with the now-unused package-registry
  AssetService wiring; deploy tests carry their tools inline in the workflow
  source closure.

## Verification

Type-check and lint pass; the non-deploy suites pass (6500+ package tests, 0
failures). The DB-backed workflow-deploy suite passes except for four
pre-existing environmental timeout flakes (real-agent and mail-timing tests)
that reproduce identically on `origin/main`. The run-lifecycle change carries an
fs/committed equivalence test plus error-path coverage for the unreadable-event
and non-ENOENT read branches; the signal validator carries reject/accept tests
including cross-body and gate-branch cases; the tarball removals are verified
against the DB-backed deploy suite and the tool-packaging end-to-end test.

Closes INTR-281
Closes INTR-482
Closes INTR-500
Closes INTR-425

### DIFF
--- a/packages/hub-sessions/src/workflow-run-kind.test.ts
+++ b/packages/hub-sessions/src/workflow-run-kind.test.ts
@@ -2970,6 +2970,24 @@ describe("claim-check API — resume-owned processing entries survive replay", (
     ).rejects.toThrow(/workflow_run_event_unreadable/);
   });
 
+  test("readWorkflowRunLifecycle surfaces a non-ENOENT events-directory read error", async () => {
+    // An absent events directory (ENOENT) classifies as absent, but a
+    // non-ENOENT read error must surface rather than be swallowed as absent.
+    // Seed the per-event `events` path as a file so readdir fails with ENOTDIR.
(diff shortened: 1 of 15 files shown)


alexanderguy:

PR #165, opened 2026-09-02
Route a unit's permanent failure to its onFailure handler step

alexanderguy:
## Summary

- The one-shot invocation primitives (`step`, `action`, `childWorkflow`) accept an
  optional `onFailure: <stepId>`. When a unit's failure is permanent — retries
  exhausted, or a crash mid-invocation — the runtime routes it to the named handler
  step instead of failing the run, following the existing route-to-handler idiom
  (`onTimeout` / `onExhausted` / `gate.else`).
- A `routed` terminal step phase marks a handled failure as distinct from a hard
  failure, so it does not count as a run failure. The routed unit's own output is a
  failure sentinel (`{ failed: true, stepId, error }`); its non-routed dependents are
  pruned, and the mirror success path prunes the handler branch.
- Definition-time validation rejects `onFailure` on primitives that cannot carry it
  and on handlers that do not `after`-depend on the unit, and rejects straddler
  handlers that cannot read the routed unit's sentinel.
- Resume reconciliation replays the prune for both terminal phases (`routed` and
  `completed`), and the projection binds `onFailure` into the approval hash.

## Verification

- `make lint` and `make build` exit 0 on HEAD.
- The workflow package suite and the deployed onFailure round-trip e2e pass (437 tests
  across the workflow package and the deploy e2e).
- The full `make all` pipeline passes on HEAD: 7302 unit tests plus the deploy
  suite, including the deployed onFailure round-trip e2e.

Closes INTR-475


alexanderguy:

PR #166, opened 2026-09-03
Flow the posix and LSP tool working directory through env-DI

alexanderguy:
## Summary

- `env.toolCwd` names the working tree the posix and LSP filesystem tools
  operate on. It is declared per tool through `defineTool`/`definePlugin`
  `requires` and is distinct from `env.workdir`, the agent's singleton lock
  and isogit storage boundary.
- The `@intx/tools-posix` and `@intx/tools-lsp` sidecar bundles read
  `toolCwd`. The production sidecar step env supplies `toolCwd` equal to
  `workdir`, and the `coding-agent` and `posix-demo` examples flow the
  working directory through env-DI rather than constructor arguments.
- The singleton lock stays keyed on `workdir` alone, so two agents may share
  one `toolCwd` while holding distinct `workdir` values.

## Verification

- `make all` passes (lint, `tsc -b --force`, admin-ui bundle, tests).
- A regression test drives a real `write_file` from each of two agents that
  share a `toolCwd` with distinct `workdir`s into the shared tree, asserts
  the singleton lock still rejects a second agent on a shared `workdir`, and
  asserts the posix bundle fails env validation when `toolCwd` is absent.

Closes INTR-170

### DIFF
--- a/apps/sidecar/src/workflow-substrate-factory.ts
+++ b/apps/sidecar/src/workflow-substrate-factory.ts
@@ -1098,14 +1098,16 @@ export function createSidecarStepBuildEnv(
     // ["capabilities", "address"]`) resolves `mail.transport` from it.
     const capabilities = createHarnessRuntimeCapabilities({ transport });
 
-    // The step env carries `transport`, `address`, and the assembled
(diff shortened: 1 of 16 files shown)


alexanderguy:

PR #167, opened 2026-09-03
Warn on cleartext sidecar transport and correct the encryption docs

alexanderguy:
## Summary

- The Git Pack Transport documentation states that hub-sidecar transport
  confidentiality depends on a TLS (wss://) deployment, rather than claiming
  the channel is already encrypted.
- The sidecar logs a warning at link creation when its hub URL is cleartext
  ws:// to a non-loopback host, because deploy frames carry decrypted
  inference keys and credential material.
- Cleartext ws:// to a loopback host stays silent for local development.

## Verification

- `make all` passes on the branch (exit 0).
- New unit tests cover the warn, no-warn (loopback and wss://), host
  normalization, and malformed-URL-throws cases.

Closes INTR-504

### DIFF
--- a/docs/IMPLEMENTATION.md
+++ b/docs/IMPLEMENTATION.md
@@ -1197,7 +1197,7 @@ Both approaches support standard git workflows for managing agent versions.
 
 ## Agent Deployment: Git Pack Transport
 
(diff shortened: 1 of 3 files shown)


alexanderguy:

One low-severity observation was raised and accepted as-is by the operator: the cleartext warning logs the full hub URL. The sidecar authenticates with a bearer token on the register frame, not URL userinfo, so no secret is exposed in the current design.

PR #168, opened 2026-09-05
Refactor: Deduplicate repeated request lifecycles and shared API plumbing

Gustav-Simonsson:
### Summary                                                                                                           
                                                                                                                       
 Seven mechanical refactors that collapse duplication into shared helpers. No behavior changes: response   
 bytes, denial strings, request-id formats, and error messages are preserved verbatim, with tests added where the      
 consolidation touches the most security-sensitive paths.                                                              
                                                                                                                       
 ### Motivation                                                                                                        
                                                                                                                       
 Auditing the duplicated request-round-trip and authorization logic kept surfacing the same two problems:              
                                                                                                                       
 1. Drift risk in security-critical gates. Bearer-token authz cross-checks and smart-HTTP principal construction were  
    copy-pasted across route groups and repo-store kinds. A fix applied to one copy could silently miss the others.    
 2. High maintenance cost for request lifecycles. The same pending-request round-trip (registration, timeout arming,   
    settling, disconnect sweep) was present five times in the sidecar handler and three times in the control-IPC   
    bridges, with each copy free to diverge on error strings and cleanup ordering.                                     
                                                                                                                       
 ### Changes                                                                                                           
                                                                                                                       
 - pending-tracker.ts — one generic websocket round-trip tracker replacing five copies in the sidecar      
   handler; per-site bookkeeping stays in call-site closures. Covered by a new pending-tracker.test.ts (198 lines).    
 - child/pending-request.ts — shared pending-request core for the three control-IPC bridges (mailbox mutation,         
   outbound mail, substrate write), keeping only wire-specific logic in each bridge.                                   
 - error-response.ts — single helper pairing each JSON error code with its HTTP status, replacing 161 inline call      
   sites across hub-api routes and middleware with byte-identical responses.                                           
 - user-principal-gate.ts (hub-sessions) — one authorization gate parameterized by verdict resource prefix, replacing  
   the four per-kind copies in agent-state, skill, workflow, and workflow-run kinds.                                   
 - git-user-principal.ts (hub-api) — shared verdict resolution / principal construction / substrate adapter glue for   
   the asset and agent-state smart-HTTP route groups.                                                                  
 - parseEventSeq — runs event filenames now parsed through the single shared authority in the workflow host            
   (supervisor compaction, cancel-signing, terminal-commit, repo-store adapter), keeping the custom compaction error   
   message verbatim.                                                                                                   
 - providers.ts (inference-catalog) — seventy provider offerings now reference shared capability-array constants per   
   model family instead of ten repeated inline literals; the re-bake guard test still verifies byte-identical values.  
                                                                                                                       
 ### Testing                                                                                                           
                                                                                                                       
 - make all — lint, tsc -b, admin-ui bundle, and full test suite pass.                                                 
 - New unit coverage for the generic pending tracker (timeout, settle, disconnect sweep paths).                        
 - Existing re-bake guard test confirms catalog offerings are unchanged after the capability-constant extraction.      
                                                                                                                       
 ### Notes                                                                                                             
                                                                                                                       
 - Pure refactor; no API, wire, or storage format changes. Exported signatures are unchanged.                          
 - Two follow-ups worth considering: the sidecar handler still carries per-site closures that could be expressed as    
   tracker options, and the error-response status table could be generated from the type definitions to keep it from   
   drifting again.


alexanderguy (approved):
LGTM; thanks for the PR

PR #169, opened 2026-09-05
Repair the forked drizzle migration snapshot chain

alexanderguy:
## Summary

- Relinks a migration snapshot to its real predecessor and restores a column the fork silently dropped from the later snapshots, so the meta chain is linear and the latest snapshot matches the applied schema again.
- Unblocks `drizzle-kit generate`: it rejected the fork as a collision, and once past that it would re-emit the dropped column as a spurious `ADD COLUMN`.

## Verification

- `drizzle-kit generate` validates the full snapshot chain and emits no spurious statements. A prevId-only fix would still re-emit `ADD COLUMN credential_refs`, which breaks `bin/db-reset` (the column is already added by an earlier migration).
- The applied SQL files and the migration journal are unchanged, so `drizzle-kit migrate`, `bin/db-reset`, and runtime are unaffected — only new-migration generation was blocked.

### DIFF
--- a/packages/db/migrations/meta/0087_snapshot.json
+++ b/packages/db/migrations/meta/0087_snapshot.json
@@ -1,6 +1,6 @@
 {
   "id": "b673a276-410f-4e34-99f9-b563273d63e1",
-  "prevId": "0f093e94-6e56-4f60-8600-53e8592155b5",
+  "prevId": "c5723916-35b0-432b-a488-4f54f49ad26f",
   "version": "7",
   "dialect": "postgresql",
   "tables": {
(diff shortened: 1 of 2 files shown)


alexanderguy:
Force-updated from the prevId-only fix to the complete repair. The fork also dropped `workflow_run.credential_refs` from the 0087/0088 snapshot contents, so relinking prevId alone left the latest snapshot drifted and `drizzle-kit generate` re-emitted that column as a spurious `ADD COLUMN` (which would break `bin/db-reset`). This revision restores the column in both affected snapshots; generate now runs with zero spurious statements. If you already pulled the earlier state, please re-pull.

alexanderguy:
Folding this repair into the INTR-164 branch, which carries the same commit and is the PR that will land it. Closing in favor of that PR.

PR #170, opened 2026-09-06
Reap the harness sidecar with a bounded kill escalation

alexanderguy:
## Summary

- Bounds the deploy-flow test teardown's sidecar reap: SIGTERM, then SIGKILL after a grace period, so the `afterAll` hook always settles.
- Removes the one unbounded `await` for a SIGTERM-deaf sidecar to exit — under teardown contention it could wedge the hook until the whole suite was torn down (~16 min), collateral-failing unrelated tests.

## Verification

- The escalation mirrors the production sidecar provisioner, which already force-kills a SIGTERM-deaf sidecar; graceful SIGTERM drain was never part of the sidecar contract.
- A full-scale `test-workflow` run with the fix in place passes 178/0 across 78 files, with teardown settling promptly on every sidecar.


alexanderguy:
Folding this fix into the INTR-164 branch, which carries the same commit and is the PR that will land it. Closing in favor of that PR.

PR #171, opened 2026-09-06
Add durable per-principal signing keys

alexanderguy:
## Summary

- Every principal (user and workflow-run) gets a durable signing key in a new `principal_key` table; the private key is sealed at rest with AES-256-GCM under a hub-held DEK (`PRINCIPAL_KEY_ENCRYPTION_KEY`), and `PrincipalKeyStore` signs internally without ever exposing private bytes.
- All principal creation routes through one owner that mints and seals the key inside the creating transaction, so no path can produce a keyless principal; a code-driven one-shot job backfills the pre-existing population (every user and workflow principal, any status).
- A discriminated `SignerIdentity` type with a store-backed lookup resolves a principal's trusted, hub-held public key for a future verifier. Hub-custodied user keys are attribution, not non-repudiation.
- Also carries two unrelated fixes the feature work surfaced: a repair of a forked drizzle migration snapshot chain that otherwise blocks generating any new migration, and a bounded sidecar-reap escalation in the deploy-flow test teardown that keeps it from wedging under contention.

## Verification

- `make all` passes on the branch: build, lint (including API-docs freshness), and all three test passes green (unit 7281, workflow 178, core 717; 0 failures).
- The backfill is idempotent and safe to run against a live hub as the last deploy step.

Closes INTR-164

### DIFF
--- a/.env.hub.example
+++ b/.env.hub.example
@@ -9,6 +9,11 @@ BETTER_AUTH_BASE_URL=http://localhost:3000
 # without it. 32 bytes, hex -- generate with `openssl rand -hex 32`.
 CREDENTIAL_ENCRYPTION_KEY=
 
+# Seals per-principal signing keys at rest. Required: the hub refuses to
+# start without it. A SEPARATE key from CREDENTIAL_ENCRYPTION_KEY so the two
+# rotate independently. 32 bytes, hex -- generate with `openssl rand -hex 32`.
+PRINCIPAL_KEY_ENCRYPTION_KEY=
+
 HUB_DATA_DIR=./tmp/hub-data
 # PORT=3000               # override the default hub port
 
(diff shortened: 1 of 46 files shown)


alexanderguy:

alexanderguy:
Added regression-guard coverage for three previously-untested paths, each verified by mutation (temporarily breaking the invariant, confirming the test fails, then restoring):

- key-mint rollback: a failed mint rolls back the principal insert in the same transaction (proves the "no keyless principal" invariant) — fails if the mint is moved outside the transaction.
- invite route: creates the principal and mints its active key — fails if the mint is a no-op.
- backfill job: hard-fails when its encryption key is absent rather than falling back to a noop cipher — fails if a noop fallback is allowed.

`make all` is green with these added (unit 7282, workflow 178, core 719; 0 failures).

PR #172, opened 2026-09-07
Sign hub-originated mail and resolve senders to durable keys

alexanderguy:
## Summary

- The workflow-run trigger — which carries both deployment triggers and user mail — signs its outbound message with the caller's hub-custodied principal key instead of a per-call throwaway key, binding each sender to a durable identity a verifier can resolve.
- A read-only resolver maps a sender's From address to the durable public key its signature must verify against (a run address to its deployment anchor's key, a user address to the sender's principal key); a `bin/` sweep audits that no signable sender is unresolvable. Nothing consumes the resolver yet.
- Tenant domains become case-insensitively unique (a `lower(domain)` unique index plus lowercase-at-creation), so a lowercased sender maps to exactly one tenant and cannot be shadowed by a case-variant registration.

## Verification

- `make all` passes: build, lint, and all three test passes green (unit 7290, workflow 178, core 736; 0 failures), including dedicated resolver, tenant-uniqueness, and trigger-guard suites.
- A whole-branch security review found no defects: no sender resolves to a wrong or cross-tenant key, ambiguous matches fail closed (never an arbitrary key), and the migration aborts rather than half-applying if legacy data collides.

Closes INTR-511

### DIFF
--- a/bin/audit-sender-keys.ts
+++ b/bin/audit-sender-keys.ts
@@ -1,0 +1,54 @@
+#!/usr/bin/env bun
+
+// Audit that every mail sender which could sign resolves to a durable public
+// key, so a durable-signature verifier sees no unresolvable signed senders.
+//
+// Read-only. Sweeps every workflow run past the deploy-ack window that carries a
+// sending address, and every active user principal, resolving each against the
+// hub's own key storage -- the same resolver the verifier uses. Run it before
+// turning on signature enforcement: it names any sender whose address does not
+// resolve so the hole is fixed before that sender's mail would be dropped. It
+// writes nothing.
+//
+//   set -a; . .env; . .env.hub; set +a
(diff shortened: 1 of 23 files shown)


alexanderguy:

alexanderguy:
Added coverage for two previously-untested paths:

- The `bin/audit-sender-keys` driver — the tool run before enabling signature enforcement. Driven as a subprocess against a seeded schema, asserting it exits non-zero on an unresolvable/running-keyless sender and refuses to run without the encryption key. Mutation-verified: replacing `process.exitCode = 1` with a no-op makes the test fail, so the exit-code contract is genuinely guarded (a silent regression here would read as a false "safe to enforce").
- The tenant-create route: the domain is lowercased at creation, and a case-variant slug returns a clean 409 (not a 500 from the `lower(domain)` index).

Test-only; `make all` green (unit 7290, workflow 178, core 740; 0 failures).

PR #173, opened 2026-09-08
Carry a hub-verified sender address on every inbound mail frame

alexanderguy:
## Summary

- Every inbound mail frame carries a required `authenticatedSender` — the
  hub-verified sender address — populated at each producer from a value the
  hub controls: the ownership-gated sender of a relayed mail, the address
  persisted at enqueue for a durable dispatch, or the triggering principal's
  address for a hub trigger. It is never taken from the spoofable MIME `From`.
- `workflow_run_dispatch` gains a `sender_address` column persisted at enqueue,
  with a status-scoped CHECK that requires a sender for deliverable mail while
  exempting hub-internal signals and terminal rows.
- The dead, spoofable `sendUserMessage` send path is removed.

## Verification

- `make all` passes (unit 7288, workflow 178, core 744; 0 failures).
- New tests cover the DB constraint (a deliverable mail row with no sender is
  rejected; a terminal one is exempt) and every producer (relay, trigger, and
  durable-dispatch frames carry the hub-verified sender, asserted distinct from
  the message's MIME `From`; the ownership gate still drops non-owned senders).
- This branch establishes the verified-sender carrier only. No recipient reads
  `authenticatedSender` yet; resolving and trusting it (retiring the MIME `From`
  read) is INTR-431, and signature verification is INTR-512.

Closes INTR-492

### DIFF
--- a/packages/db/migrations/0091_workflow_run_dispatch_sender_address.sql
+++ b/packages/db/migrations/0091_workflow_run_dispatch_sender_address.sql
@@ -1,0 +1,21 @@
+ALTER TABLE "workflow_run_dispatch" ADD COLUMN "sender_address" text;--> statement-breakpoint
+-- Fail in-flight mail dispatches that predate the authenticated-sender
+-- requirement. They carry no persisted hub-verified sender, and neither
+-- fabricating one nor re-reading the signed From is permitted, so they are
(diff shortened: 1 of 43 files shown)


alexanderguy:

PR #174, opened 2026-09-09
Pin non-agent steps to an inert placeholder and withhold their credentials

brianjfox:
## Summary

- A workflow step that cannot invoke inference — `action`, `gate`, `awaitSignal`, `sleep`, `loop`, `escalation`, `onTrigger`, `childWorkflow` — is pinned the deploy's default source as an inert placeholder and is not approval-gated. The pin exists because the wire shape requires a source for every step, not because it names a route the step takes. Agent steps resolve through `pickStepInferenceSource` and its operator-approval gate.
- The hub delivers an inference credential only for a step that can issue a request. Which steps those are is read from the hash-covered projection rather than from the pinned sources map, which carries an entry for every step and so cannot answer the question.
- The rule that resolves a non-agent leaf lives in one place, `@intx/workflow-deploy`, beside the walk and the resolver that own pin policy. Both the top-level pin and the lifted-body pin use it.

## Why

A code-sourced workflow made only of deterministic primitives cannot deploy through `POST /api/tenants/:tenantId/workflows/deployments`:

```
409 invalid_workflow: workflow definition "…" is invalid: step <firstStep> has no approved inference
source: the step declared no preferred source is either missing from HarnessConfig.sources or not in
the operator-approved grant set, and the deploy's defaultSource "mof_…" does not resolve to an
operator-approved source
```

- The probe's advertised grant set comes from the capability walk, and `inference.source:<provider>:<model>` grants are emitted only from agent definitions (`packages/workflow-deploy/src/capability-walk.ts`).
- The route installs with `approvals: { mode: "approve-probed" }`, so the approved set equals the advertised set.
- Pinning every top-level step through the approval gate therefore demands a grant that a definition with no agent step can never advertise.
- `validateWorkflowProjection` requires a non-empty `sources[stepId]` for every step, so the pin cannot be skipped.

The gate's reach is wider than agent-free definitions: it admits a workflow containing any non-agent step only when the tenant's default source carries a `(provider, model)` that some agent in the same workflow also declares.

Delivering a credential for a step pinned to a source it never uses decrypts tenant material, puts it on the deploy frame, seals it to sidecar disk, and re-resolves it on every reconnect with no re-authorization.

## Reproduction

```ts
import { action, defineWorkflow } from "@intx/workflow/definition";
export const workflow = defineWorkflow({
  id: "wf_all_action",
  trigger: { type: "manual" },
  steps: { gather: action({ handler: "gather", effect: { requires: ["domain:read"] } }) },
});
export async function gather(input) { return { input }; }
```

Publish it as a `workflow`-kind asset in source format and deploy it with a valid `sourceOfferingIds` / `defaultSourceOfferingId` chain. The probe succeeds and the gate approves `effect:domain:read`; the source pin then throws for step `gather`.

## Verification

- `make all` passes on each commit independently, including the integration passes.
- `tests/workflow-deploy/all-action-deploy.test.ts` deploys a workflow with no agent step against a real hub and sidecar and asserts both its actions run.
- Credential delivery is asserted on both rails, top level and lifted body, each with a distinct credential so the exclusion is observable.
- `collectAgentBearingStepIds` is covered for `map`, for `loop` containers against their bodies, for `onTrigger` containers, for nested loops, and for an agent that declares no model sources.
- The placeholder resolves per leaf, so a definition whose steps all carry their own source deploys with a dangling `defaultSource`.

Marking a placeholder inert on the wire is out of scope here: it reaches the sidecar source-admission gate, the body-sources size ceiling, and the wire source shape. It is tracked separately.


packages/workflow-deploy/src/orchestrator.ts
```diff
@@ -227,14 +227,35 @@ export function buildInertProjectionStepSources(args: {
...
+      // escalation, onTrigger, childWorkflow) never invokes inference, so
+      // its pinned source is an inert placeholder: the deploy's default
+      // source, exactly as `buildReferencedWorkflowSourcePins` pins a
+      // non-agent BODY step. Routing it through `pickStepInferenceSource`
+      // gated it on an `inference.source:` approval the capability walk
+      // only ever emits for agent definitions, which made a definition
+      // with no agent step -- an all-action workflow -- undeployable
+      // under `approve-probed`. Agent steps keep the full resolver and
+      // the operator-approval gate.
+      if (!isAgent) {
```
alexanderguy: The doc block above this function describes the behavior it now has: an agent-bearing leaf resolves through `pickStepInferenceSource` and its operator-approval gate, and a leaf that cannot invoke inference takes the inert placeholder ungated. The module header carries the same distinction, so neither claims the pin is approval-gated for every step.

packages/workflow-deploy/src/orchestrator.ts
```diff
@@ -227,14 +227,35 @@ export function buildInertProjectionStepSources(args: {
...
+        const placeholder = args.config.sources.find(
+          (source) => source.id === args.config.defaultSource,
+        );
+        if (placeholder !== undefined) return placeholder;
+        throw new WorkflowDefinitionInvalidError(
+          args.projection.id,
+          `non-agent step ${stepId} needs an inert placeholder source, but the deploy config carries no defaultSource entry to pin`,
+        );
```
alexanderguy: The non-agent leaf rule has a single home: `inertPlaceholderSource` in `@intx/workflow-deploy`, reached by the top-level pin here and by `buildInertBodyStepSources` for a lifted body, which `buildReferencedWorkflowSourcePins` in `hub-sessions` calls. One `find default -> else throw`, one error message, no second copy to drift.

alexanderguy (commented):
**Automated review — this feedback was produced by a pull-request-review agent, not a human reviewer.**

Tightly-scoped change, and the diagnosis holds. `buildInertProjectionStepSources` routes a leaf that cannot invoke inference to the inert placeholder and keeps the picker plus its approval gate for agent leaves, so an all-primitive workflow deploys without the gate demanding a grant the capability walk only ever emits from agent definitions. The leaf rule has one home in `@intx/workflow-deploy` and both pins use it, so the top-level and body rails cannot drift apart again.

The credential half is the part worth reading closely. `prepareSourceRefDeploy` derives the entitled step set from the hash-covered projection rather than the pinned map, so a step that fills the wire's per-step coverage slot contributes no `credentialId`, and no tenant secret is decrypted, framed, sealed to sidecar disk, or re-resolved on reconnect on its behalf. Deriving it from the projection rather than accepting it as an argument keeps that decision off the caller's surface.

Verified rather than eyeballed:

- `make all` passes on each commit independently, integration passes included.
- The un-gated branch is reachable only for genuine non-agent primitives: `readInertStepInference` throws on a `step`/`map` lacking valid `agent.modelSources`, so a malformed agent step cannot slip through as "non-agent", and an agent declaring empty `modelSources` stays gated.
- `tests/workflow-deploy/all-action-deploy.test.ts` deploys an agent-free workflow against a real hub and sidecar and asserts its actions run.
- Credential delivery is asserted on both rails with distinct credentials, so the exclusion is observable rather than collapsing to one row.

Two notes, neither blocking. `pinInertStepSources` is now exported but reached only from within its own module, so the published surface is wider than its use. And `readInertStepInference` is not exhaustive over step kinds — it answers "not agent-bearing" for anything it does not recognize, which is fail-open for a primitive added later that can invoke inference.

PR #175, opened 2026-09-09
Shadow-verify inbound mail signatures at the sidecar ingress

alexanderguy:
## Summary

- Inbound mail is signature-verified recipient-side at the sidecar's
  `mail.inbound` ingress, in SHADOW mode: the verdict is logged and every
  message is admitted — nothing is rejected yet.
- The check is keyed on the hub-stamped `authenticatedSender` (never the
  spoofable MIME `From`), against a hub-resolved sender key delivered on the
  frame. It logs a two-axis verdict — `signature` (valid|invalid|missing|
  unknown|error) and `fromMatch` (match|mismatch|unchecked) — so a valid
  signature carried under a forged `From` is caught as a mismatch.
- A shared MIME signature-verify helper is extracted for reuse. The frame
  carries the hub-resolved key (nullable), resolved via a helper that degrades
  a resolver throw to a null key plus a distinct ERROR log, so a shadow-only,
  verification-optional field can never block delivery on any producer path.

## Verification

- `make all` passes on the implementation (unit + workflow-deploy + core, 0
  failures); the final comment-only amend passed build + lint; CI re-runs the
  full suite on this PR.
- Verified end-to-end at the real hub-link `mail.inbound` WS ingress: a
  validly-signed frame logs `valid/match`, a tampered signature logs `invalid`,
  an unresolvable sender logs `unknown`, and a valid signature under a forged
  `From` logs `valid/mismatch` — all admitted (shadow).
- The new `hub-agent` -> `@intx/mailbox`/`@intx/mime` dependency is acyclic.

This is the SHADOW stage of INTR-512 — it logs and admits, never rejects. The
enforce flip (reject on bad/absent signature, with a drain-before-flip) is a
follow-up PR on the same issue, gated on a clean shadow observation.

Part of INTR-512

### DIFF
--- a/apps/hub/src/server.ts
+++ b/apps/hub/src/server.ts
@@ -4,6 +4,7 @@ import {
   createPrincipalKeyStore,
   createSidecarAllocationStore,
   createWorkflowRunDispatchStore,
+  resolveFrameSenderKey,
 } from "@intx/db";
 import { createEnvKeyCredentialCipher } from "@intx/crypto";
 import { hexDecode, type SidecarCapabilityRule } from "@intx/types";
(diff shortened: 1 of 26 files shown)


alexanderguy:

PR #176, opened 2026-09-09
INTR-503 - Stop retrying permanent sidecar destroy rejections

FarinM:
# Stop retrying permanent sidecar destroy rejections

## Why

The reconciler ignored `retryable=false` on destroy rejections, leaving allocations stuck in `releasing` or `replacing` and retrying indefinitely.

## What this adds

Add a terminal `destroy_failed` state that records the provider failure, fails remaining live runs and unsettled dispatches, and stops automatic reconciliation. Preserve infrastructure references and the sidecar reservation for operator cleanup, and expose the state in deployment responses.

Keep retrying transient rejections and thrown errors. Cover both release and replacement failures, stale reconciliation attempts, and reservation preservation with reconciler, database, and API tests.


alexanderguy (approved):
LGTM

PR #177, opened 2026-09-09
Open the hub database to any pg driver and drop hub-agent's dead deps

alexanderguy:
## Summary

- Remove two dead dependencies from the published `@intx/hub-agent`: an
  unused `@intx/test-harness` devDependency (it points at the `tests/` tree,
  so it breaks `bun install` for a consumer that vendors only `packages/*`)
  and an unused `@intx/harness` runtime dependency.
- Type the hub database against the driver-agnostic `PgDatabase` base
  intersected with a required `$client`, so the store factories accept a
  database built on any drizzle pg driver — a caller-built pglite database
  included — while a bare transaction is still rejected. `createDB` still
  returns the concrete postgres-js handle, and `@intx/db` gains no pglite
  dependency.

## Verification

- `make all` passes (build, lint, full-graph type-check, and tests exit clean).
- A `packages/*`-only vendored subset resolves `bun install` without the
  `tests/` tree, since the dependency pointing there is gone.
- Type-level: a pglite database from `drizzle(handle, { schema })` is
  assignable to the store-facing type; a transaction is not. Runtime execution
  of the stores over pglite is not exercised here — that needs a pglite
  dependency this branch deliberately omits.

Closes INTR-521

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -330,7 +330,6 @@
       "devDependencies": {
         "@intx/crypto": "workspace:*",
         "@intx/hub-sessions": "workspace:*",
-        "@intx/test-harness": "workspace:*",
         "hono": "catalog:",
       },
     },
--- a/bun.lock
+++ b/bun.lock
@@ -316,7 +316,6 @@
       "name": "@intx/hub-agent",
       "version": "0.3.0",
       "dependencies": {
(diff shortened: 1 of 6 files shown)


alexanderguy:

PR #179, opened 2026-09-10
Verify inbound mail against a co-delivered sender-key cache

alexanderguy:
## Summary

- The hub co-delivers each authorized sender's resolved public key on the same
  `run.grants` barrier as the authorization grant (via a required
  `sendRunGrants` parameter, so no push site can silently drop it), and the
  sidecar caches each key in a local disk-backed keyring — written before the
  run's grants file so a durable grant always implies a durable key, and
  eager-loaded at boot so verification survives a restart or disconnection.
- A resolver wraps a cached key as a public-key-only crypto provider matching
  the inbound-mail verify seam.
- The recipient-side inbound-mail signature check verifies each message against
  the cache-resolved sender key and logs the verdict. It is shadow-only: it
  never rejects mail. The interim key that rode on the `mail.inbound` frame is
  removed — the cache is now the recipient's key source.

## Verification

- Lint, `tsc -b --noEmit --force`, the admin-ui production bundle, and the full
  unit suite (7354 tests) pass.
- The recipient verify is covered by unit and real-WS-ingress tests with real
  crypto — valid / invalid / missing / `unknown`-on-cache-miss and both
  fault-containment paths; the cache's write-before-grants durability,
  malformed-key skip, and the co-delivery paths have dedicated tests.
- The real-process `test-workflow` end-to-end suite runs in CI as the
  authoritative integration gate.
- The verify path is fault-contained: a resolver, `getPublicKey`, or verify
  fault degrades to an `error` verdict and never crashes the mail path. A cache
  miss logs `unknown` and admits — expected for grants that predate the cache,
  relayed agent-to-agent mail with no preceding co-delivery, and the
  not-yet-built reconnect key-refresh.

Closes INTR-517
Closes INTR-519

### DIFF
--- a/apps/sidecar/src/agent-key-registration-lifecycle.test.ts
+++ b/apps/sidecar/src/agent-key-registration-lifecycle.test.ts
@@ -252,6 +252,7 @@ describe("agent signing-key registration lifecycle on the host transport", () =>
       } as unknown as Parameters<
         typeof createSidecarDeployRouter
       >[0]["keyStore"],
+      senderKeyCache: { get: () => undefined, put: async () => undefined },
       transport,
       repoStore,
       signingKeySeed: keyPair.privateKey,
--- a/apps/sidecar/src/index.ts
+++ b/apps/sidecar/src/index.ts
@@ -10,6 +10,7 @@ import {
 } from "@intx/crypto";
 import {
   createSenderKeyCache,
+  createSenderCryptoResolver,
   createSidecarOrchestrator,
   type HubLink,
 } from "@intx/hub-agent";
@@ -245,6 +246,12 @@ const senderKeyCache = await createSenderKeyCache({
     writeFileAtomicDurable(filePath, contents, { mode: 0o600 }),
 });
 
+// The read side of the same cache: the inbound signature shadow resolves a
+// sender address to the crypto that verifies its mail. Built here at the edge
+// so the hub link stays source-opaque -- it resolves address -> crypto without
+// holding the cache or knowing where the key came from.
+const resolveSenderCrypto = createSenderCryptoResolver(senderKeyCache);
+
 // The deploy router records `(runId -> agentAddress)` here on
 // every inbound `agent.deploy`; the facade resolves the mapping when
 // firing the pack push so the outbound frames carry the right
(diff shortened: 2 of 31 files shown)


alexanderguy:

- `packages/hub-agent/src/sender-crypto.ts`: `createPublicKeyCrypto().verify()`
  throws a directive error instead of verifying. Greybeard waiver: the
  `CryptoProvider.verify` seam verifies over raw bytes, but an inbound mail
  signature is over the canonicalized MIME representation
  (`verifyMimeSignature`'s job), so the throw steers a future caller off a
  silently-wrong path toward the correct one; implementing it would also add
  `hub-agent`'s first production `@intx/crypto` dependency for a method no
  seam consumer calls.

alexanderguy:
Force-pushed an update (`ce5ca063` → `9045c982`): the recipient-side check now
verifies inbound mail directly against the cache-resolved sender key, and the
interim frame-delivered key (`authenticatedSenderPublicKey` on `mail.inbound`,
its producer stamps, and the `routeMail` key param) is removed. The earlier
compare-against-the-frame-key step is gone — the cache is the recipient's key
source. This folds in what was tracked as INTR-519.

Still shadow (log + admit); enforce remains a later flip, now gated on the
cache-sourced verdict being observed clean. The greybeard waiver below still
applies (the verify-only provider's `verify()` throws by design).

Whole-branch `make all` green on the reworked branch (`MAKE_EXIT=0`;
test-workflow drives the cache-sourced verify end-to-end).

alexanderguy:
Force-pushed a small follow-up (`9045c982` → `e133950c`): a single-hunk change
in `deliverMailToRecipient` that resolves the sender key inside the
`materialized` run-grants branch instead of unconditionally at the top, so a
non-run relay path, a `skip`, or a rejected materialization no longer pays for
a lookup it would discard. Behavior is identical for run recipients (every
grant-sending delivery hits the `materialized` branch); the durable-dispatch
path is untouched.

Lint, tsc, the admin-ui bundle, and the full unit suite pass locally; the
`test-workflow` e2e is left to CI as the authoritative gate (this change is off
that suite's surface, and the prior tree passed it in full).

PR #180, opened 2026-09-10
INTR - 350 - Reduce validator memory by constructing unions once

FarinM:
Long ArkType `.or()` chains compile intermediate unions during module initialization, increasing startup memory. Replace ten chains with single `type.or(...)` calls across shared types, workflow control messages, Anthropic streaming events, and discovery response formats.

The rewrite preserves branch definitions, tool-result array validation, and the manual `InferenceEvent` type used for `custom.*` narrowing. JIT validation remains enabled.

Fresh-process import measurements show approximately 166–176 MiB lower RSS on the workflow-child and sidecar wiring paths:

| Import path    |    Before |     After | Reduction |
| -------------- | --------: | --------: | --------: |
| Workflow child | 510.4 MiB | 344.5 MiB | 165.9 MiB |
| Sidecar wiring | 532.3 MiB | 356.3 MiB | 176.0 MiB |

Compared with the shared-types-only version measured in the same run, the additional IPC, Anthropic, and discovery rewrites saved another 25.2 MiB on the workflow-child path and 23.6 MiB on sidecar wiring.

Measurements are medians of three fresh-process runs per variant on macOS arm64 with Bun 1.3.9 and ArkType 2.2.0. They cover module imports.

Validation:

- `make all` passed: 8,252 tests passed, with two live API tests skipped.
- The focused suites passed: 162 tests for shared types and 117 for IPC, Anthropic, and discovery. Generated API documentation passed its freshness check.
- Before/after comparisons preserved inferred input/output types and exported schemas, including metadata and supported JSON Schema output.

### DIFF
--- a/packages/inference-discovery/src/catalog/intent.ts
+++ b/packages/inference-discovery/src/catalog/intent.ts
@@ -44,18 +44,20 @@ const FollowUp = type({
 // response_format, Gemini's responseSchema; Anthropic does not
 // receive a structured-output request because its adapter rejects
 // the field at the marshaling boundary).
-const ResponseFormatIntent = type({
-  kind: "'text'",
-})
-  .or({
+const ResponseFormatIntent = type.or(
+  {
+    kind: "'text'",
+  },
+  {
     kind: "'json'",
-  })
-  .or({
+  },
+  {
     kind: "'json-schema'",
     name: "string",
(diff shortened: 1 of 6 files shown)


alexanderguy (approved):
LGTM

PR #181, opened 2026-09-10
Run workflow-deploy tests concurrently, with a configurable worker count

Gustav-Simonsson:
## What

The workflow-deploy integration pass was the bottleneck in `make test`: ~81% of full-suite wall clock (624s of ~766s)

This while averaging under 1 core. Each test file is wall-clock bound by sidecar boot and reconnect waits.

Files are fully isolated (port 0, `mkdtemp` data dirs, one unique Postgres schema per file), so they are safe to run concurrently.

This PR runs the pass with `bun test --parallel=4` and makes the worker count configurable via `WF_PARALLEL`.

## Changes

- **Makefile `test-workflow`**: run the suite 4-way in parallel (`--parallel=4`).
- **Hang-detection budget**: the parallel load exposed a 6ms wall-clock overrun in the inference harness's default 250ms per-drain budget. Widen it to 2000ms at the four `harness.run()` call sites in the workflow-deploy suite (`wallClockBudgetMs: 2000`). No assertions or coverage change; only headroom for scheduling jitter.
- **Makefile `WF_PARALLEL`**: new variable (default 4), passed as `--parallel=$(WF_PARALLEL)`. Bigger machines can raise it without editing the Makefile: `make test-workflow WF_PARALLEL=8`. Default stays 4 for the 16GB/4-vCPU CI runner.

## Metrics (vs single worker)

| | 1 worker | 4 workers (default) |
|---|---|---|
| workflow-deploy pass | ~624s | ~230–286s (**~2.2–2.7x**) |
| full `make test` | ~13 min | ~7.5–8 min (**~1.6x**) |

Measured on an 8-core/31GB guest; no change in flakiness at 4-way or 8-way.


alexanderguy (approved):
Note up front: this is an agentic review, produced with the
`pull-request-review` skill.

I spent most of my time trying to poke holes in the isolation argument, since
that is what the whole change rests on, and it held up. Every listener binds
`port: 0`, the temp dirs are all `mkdtemp`, and `randomSchemaName` draws from
`Math.random()` rather than a counter or a pid, so the names stay unique once
the files are spread across separate worker processes — that last one is the
detail I expected to be the problem, and it isn't. Worth noting the one place
it could have gone wrong and didn't: `listWorkflowHostChildren` shells out to a
system-wide `ps -A`, but walks the tree from its own sidecar's pid rather than
pattern-matching on argv, so it can't reach a sibling file's children.

The default of 4 was green across every local run I did, and CI is green.

One small thing, plus an aside.

## `CONVENTIONS.md` line 714 reads oddly now

The Testing section still opens with "Tests that are not parallel-safe (spawn
servers, perform real `isomorphic-git` operations against `os.tmpdir()`, or
otherwise need extended timeouts) run in separate passes from the fast unit
suite." With this change, someone reading top to bottom meets a pass described
as not parallel-safe and then finds it running under `--parallel`.

The original meaning — these can't be folded into the fast unit pass — is still
accurate. It just now needs distinguishing from "can't run concurrently with
each other," which this PR establishes they can. Maybe something like:

> Tests that cannot run inside the fast unit pass (they spawn servers, perform
> real `isomorphic-git` operations against `os.tmpdir()`, or otherwise need
> extended timeouts) run in separate passes. Within a pass, files may still run
> concurrently where they are isolated from one another.

Entirely your call on wording — I mainly wanted to flag that the sentence and
the Makefile now point in different directions.

## Aside: the deploy budget in `deploy-flow-env.ts`

Not for this PR, and not something I have good numbers on. While reading
through, I noticed `requestTimeoutMs: 10_000` there against
`DEFAULT_REQUEST_TIMEOUT_MS = 30_000` in `sidecar-handler.ts` — so the harness
tightens the deploy budget to a third of the production value, and 70 of the 79
files in the pass inherit it. That tightening looks like it predates any
parallelism.

It's the budget I'd expect to be the first to complain on slower or more
loaded hardware, in the same way the 250ms drain budget did here. Might be
worth a look at some point, independently of anything in this PR.

Gustav-Simonsson:
Added a new commit updating CONVENTIONS.md as per the review suggestions.

alexanderguy:
LGTM, thanks for the update.  Feel free to merge.

PR #182, opened 2026-09-11
Refresh cached sender keys on sidecar reconnect

alexanderguy:
## Summary

- On (re)connect the sidecar reports the sender addresses it has cached; the
  hub re-resolves each sender's current key and pushes the refreshed key back,
  so a user-principal key that rotated while the sidecar was disconnected no
  longer leaves a stale cached key (which would reject legitimate mail once
  verification enforces).
- The refresh rides a dedicated key-only frame that updates the sidecar's cache
  via `SenderKeyCache.put` alone — no grants rewrite, so a transient write fault
  logs at ERROR but never poisons a run.
- Only rotatable (user) senders are reported; the hub re-resolves every reported
  address itself rather than trusting the sidecar's filter, and caps the set it
  will resync so a hostile sidecar cannot drive unbounded resolution.

## Verification

- Lint, `tsc -b --noEmit --force`, the admin-ui bundle, and the full unit suite
  pass. The new behavior is covered by unit and integration tests that run in
  the unit suite: the key-only refresh handler (fault-swallowed, no poison,
  wrong-length rejected), the resolve-don't-trust hub resync across both the
  register and reconnect paths, the resync cap (over-cap truncates and warns,
  under-cap refreshes all), and the rotated-key reconnect end to end — a real
  `HubLink` + real `SenderKeyCache` + a cold-load-from-disk round trip.
- The real-process `test-workflow` end-to-end suite (the workflow-deploy
  reconnect tests) runs in CI as the authoritative end-to-end gate.

Closes INTR-520

### DIFF
--- a/apps/sidecar/src/agent-key-registration-lifecycle.test.ts
+++ b/apps/sidecar/src/agent-key-registration-lifecycle.test.ts
@@ -252,7 +252,11 @@ describe("agent signing-key registration lifecycle on the host transport", () =>
       } as unknown as Parameters<
         typeof createSidecarDeployRouter
       >[0]["keyStore"],
-      senderKeyCache: { get: () => undefined, put: async () => undefined },
+      senderKeyCache: {
+        get: () => undefined,
+        put: async () => undefined,
+        addresses: () => [],
+      },
       transport,
       repoStore,
       signingKeySeed: keyPair.privateKey,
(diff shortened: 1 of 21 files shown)


alexanderguy:

The whole-branch review's one observation — the unbounded reported-address
set — is *closed* by the resync cap (`MAX_RESYNC_SENDER_ADDRESSES`, final
commit), not waived. A related pre-existing platform concern it surfaced
(unbounded `string[]` frame fields, and the control socket's whole-frame
payload model — neither introduced by this branch) is tracked as follow-ups
INTR-525 and INTR-526; INTR-526 also captures a latent >16MB frame-truncation
bug the investigation exposed.

PR #183, opened 2026-09-11
Harden the sidecar control-socket frames and sender-key cache

alexanderguy:
## Summary

- **Bound the unbounded `string[]` frame fields** (INTR-525): fail-loud
  `.atMostLength` limits on `agentAddresses`, `cachedSenderAddresses`, mail
  `recipients`/`to`/`cc`, probe `grants`, and credential `revoke`, so an
  oversized wire frame is rejected at the parse boundary. The
  `cachedSenderAddresses` ceiling sits well above the reconnect resync handler
  cap, so a merely-large report still hits that graceful degrade rather than a
  hard reject.
- **Pin the frame payload limits** (INTR-526): an explicit websocket
  `maxPayloadLength` — driven by the 44 MB received `mail.outbound` frame,
  which fixes a latent truncation from Bun's 16 MB default — plus an app-layer
  byte cap on `mail.outbound` (hub-receive warn+drop as the authoritative
  backstop, send-site throw to surface to the producer, audit-site log),
  symmetric with the inbound mail-body cap.
- **Evict a deleted sender's cached key on reconnect** (INTR-524): on reconnect
  the hub re-resolves each reported cached sender and, when the sender is
  *confirmed absent* (hard-deleted), pushes a `sender.key.evict` that durably
  removes the key from the sidecar cache. A resolver *fault* keeps the stale
  key rather than dropping a live one — a strict tri-state (resolve→refresh,
  confirmed-absent→evict, throw→keep). Hard-delete only by design.

## Verification

- Lint, `tsc -b --noEmit --force`, the admin-ui bundle, and the full unit suite
  pass (7408 tests, 0 fail). The new behavior is covered by unit and
  integration tests: the frame-bound rejections and the ceiling > resync-cap
  invariant; the mail-body cap (over-cap drop, send-throw, audit-log) and the
  `maxPayloadLength` > mail-cap guard; the durable evict; and the load-bearing
  tri-state case — a resolver fault does NOT evict.
- The real-process `test-workflow` end-to-end suite (the reconnect suites
  especially, which exercise the resync loop INTR-524 touches) runs in CI as
  the authoritative end-to-end gate.

Closes INTR-524
Closes INTR-525
Closes INTR-526

### DIFF
--- a/apps/hub/src/server.ts
+++ b/apps/hub/src/server.ts
@@ -35,6 +35,7 @@ import {
 } from "@intx/hub-sessions";
 import { generateKeyPair } from "@intx/crypto";
 import { hexEncode } from "@intx/types";
+import { MAX_SIDECAR_FRAME_BYTES } from "@intx/types/sidecar";
 import { upgradeWebSocket, websocket } from "hono/bun";
 import { setup, getLogger } from "@intx/log";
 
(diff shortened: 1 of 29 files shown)


alexanderguy:
Self-review (`code-review` skill) returned clean — no defects across the three
commits.

The whole-branch review's one observation (a spurious cold-cache
parent-dir-fsync WARN on evict) is a non-defect, not a waiver: a sidecar only
reports cached senders it actually holds, so the sender-keys directory always
exists when an evict runs, and the WARN cannot occur on the real path (and it
degrades, never throws, regardless). No code change and no waivers.

alexanderguy:
Force-pushed: added a full reconnect->evict integration test (real HubLink + real SenderKeyCache, cold-loaded from disk) asserting a hard-deleted sender's cached key is removed on reconnect. Mutation-verified (dropping the hub-side evict send fails only this test) and critique-clean. Amended into the evict commit; the other two commits are unchanged. Also re-triggers CI (the prior run was stuck pending on runner/queue infra).

PR #184, opened 2026-09-12
Enforce inbound mail admission with a per-workflow signature policy

alexanderguy:
## Summary

- Recipient-side inbound mail is admitted per an author-declared, content-hashed `inboundMailPolicy` on the workflow definition: the signature verdict reduces to an admission outcome, and the recipient's policy decides admit-or-drop at the delivery seam. The secure default rejects anything not `valid` and verified; an unauthored policy and an unregistered address both fail closed.
- Two windows that produced a spurious `unknown` verdict for legitimate mail are closed: a pre-ack run holds its outbound mail until its key is recorded, and keyless pending-mail replay re-resolves a run sender's key.

## Verification

- `make all` passes: lint, `tsc -b --force`, the admin-ui bundle, and the unit/integration suites.
- Every commit went through architectural and code review; the two concurrency changes passed dedicated correctness review.
- A real-route end-to-end test exercises a deployed workflow admitting valid mail, dropping unsigned mail under the default policy, and admitting the same unsigned mail under a policy that relaxes it. The `test-workflow` e2e suite runs in CI.

Closes INTR-512


alexanderguy:

PR #185, opened 2026-09-14
INTR - 536 Unrelated workflow deployments and recovery can stall during provisioning

FarinM:
# Unrelated workflow deployments and recovery no longer stall on slow provisioners

## Why

One slow or hung provisioner call (ensure, destroy, or workflow initialization) held the single allocation reconciliation loop, so every unrelated pending deployment queued behind it. The Hub also had no way to cancel in-flight work: a disconnect, lease loss, or restart left the old attempt running with no owner, and its late completion could publish a stale key or clear the next attempt's sender-key marker. Startup additionally awaited provider destruction for interrupted probes, so an unavailable provisioner held the Hub startup path.

This change bounds every provider and initialization wait, cancels stale attempts via the existing lease and connection events, defers destruction out of startup, and reconciles independent allocations on separate slots.

## What this adds

- Bounded provider and initialization waits (120s default, `sidecarOperationTimeoutMs` / `operationTimeoutMs` configurable). The Hub passes an abort signal to `ensure`/`destroy` and stops waiting at the deadline even if the provisioner ignores cancellation. Ensure timeouts are treated as uncertain (fence and replace); destroy timeouts leave cleanup pending and never authorize replacement. Late completions must still respect the generation fence.
- Lease- and connection-driven cancellation. In-flight reconciliations track an abort controller per allocation; disconnect, lease-expiry, and timeout abort the attempt. Final writes go through a per-allocation event queue so a racing connect/disconnect is not lost, and a connect during initialization forces another readiness pass. DB lifecycle transitions now require the lease to be current (id match plus unexpired), so expired owners fail closed.
- Cancellation plumbed through initialization. `deployReadyAllocation` carries the reconciliation context into restore packs, deploy frames, and the anchor-key publish; checks before and after each step prevent sends and key writes after cancellation. The sender-key settle is abort-aware, so a cancelled attempt settles failed and cannot clear the next attempt's marker.
- Probe cleanup deferred out of startup. Startup only marks interrupted probes `releasing`; provider destruction runs in the independent 30s cleanup scheduler. Callers in one Hub share one in-flight destroy per probe and reread state before destroying; `releasing` stays the durable retry obligation.
- Concurrent reconciliation. A small generic scheduler replaces the single loop: allocation (8 slots by default, `sidecarAllocationConcurrency`), probe cleanup, dispatch, and connection repair run on independent schedules. Slots claim disjoint work and a slow slot does not block the others. Dispatch polls enter the same guarded drain as enqueue notifications.


alexanderguy (changes requested):
I'm an automated pull-request-review agent, not a human reviewer. I ran the full build,
lint and test gate against a3dd2d1e and it all passes, so nothing below is a test failure.
Everything here I checked against the code directly. If I've misread something I'd rather
be corrected than deferred to.

The per-allocation event queue is the part of this I'd single out. The identity-checked
cleanup and the previous.catch that keeps a thrown task from stalling the chain are both
easy to get wrong, and neither is. It holds up under concurrent connect and disconnect
events for the same allocation: nothing drops, nothing reorders.

Two things I think should be resolved before merge.

The first is that an initialization overrun now permanently fails the deployment. onReady
runs inside runSidecarOperation("Workflow initialization", 120_000) at reconciler.ts:376.
A SidecarOperationTimeoutError there goes to replaceAfterFailure with the code
sidecar_initialization_uncertain (reconciler.ts:385). Because
enableAutomaticReplacementRecovery defaults to false at reconciler.ts:148 and is never set
outside reconciler.test.ts:731, that lands on the unrecoverable branch at reconciler.ts:303,
which calls beginUnrecoverableRelease. That fails every live run and deactivates their
principals (sidecar-allocation-store.ts:317-325).

The concern is that the work inside that 120s cap can legitimately need longer. A restore
sends two refs sequentially, each bounded by PACK_TIMEOUT_MS, which is requestTimeoutMs * 4
= 120s (sidecar-handler.ts:2711 and :495), and then a deploy frame at 30s. A cold first
deploy never gets near it, because restoreWorkflowRunToAllocation skips refs that resolve
to null, so this only bites a replacement restoring a run that has history. That is the
recovery path this PR is about. On main there was no deadline on onReady at all, so a slow
restore threw a plain Error and took the scheduleRetry branch with backoff. Your test
"fences initialization when its deadline expires" asserts this route, so the timeout is
clearly deliberate; my question is whether the permanence is. Would it make sense to give
initialization its own budget rather than sharing the provider-call constant, or to send an
init timeout to the retry branch and keep uncertain/replace for actual provider calls?

The second is that an initialization failure can lose its failure record. noteConnect at
reconciler.ts:788 sets pendingConnect whenever a connect arrives for an allocation this Hub
is mid-claim on, and nothing ever consumes it. So the pendingConnect branch in
finishReconciliation at reconciler.ts:193 fires for the connect that enabled the pass, not
only for a later reconnect. That branch doesn't look at which write it was handed, so it
replaces the initialization-failure scheduleRetry at reconciler.ts:400 wholesale. With a
connect during init the retry is recorded with failure undefined and nextAttemptAt equal to
now. Without one, the same failure records code sidecar_initialization_failed at now plus
30s. I reproduced both.

This isn't a rare collision between a failure and a reconnect. The trigger is any pass this
Hub was holding when the worker registered, which is the ordinary provisioning sequence. The lost
failureCode is the durable damage, since the allocation then retries with nothing recorded
and the UI shows it retrying for no stated reason. The missing backoff is real but bounded,
because the next claim starts with pendingConnect null and does record the failure.

The same state also costs a redundant pass on the success path: the branch schedules a retry
instead of markConnectionReady, giving initialize, retry, initialize, ready. That second pass
is cheap, since deployReadyAllocation returns early at workflow-allocation-service.ts:719-720
when the workflow is active and the anchor key is set, so it is one extra claim and one
write rather than a second deploy. Both come from the same fix: consume pendingConnect when
the reconciler acts on the connect it was waiting for. Note that the existing test
"preserves a connect received while a newly bound generation initializes" asserts the
doubling as intended, so it would need re-expressing.

Three more that I don't think block, but are worth a look.

Lease loss is now unlogged, and any thrown error cancels a valid deploy. The catch in renew()
at reconciler.ts:254 turns any throw from extendReconciliationLease into a lease-lost abort.
That's a plain drizzle UPDATE over postgres.js, so a closed connection or query timeout
rejects rather than returning false, and a transient DB fault will cancel an in-flight deploy
whose lease is still valid. It's also silent: this branch deletes main's warn for a failed
extend, and the return at reconciler.ts:871 sits above the logger.error. The row then stays
leased until the last successful renew plus 60s. Worth separating "the store said no" from
"we couldn't reach the store", and logging either way?

Cancellation doesn't reach the two waits that take time. signal is checked twice in
sendWorkflowRunPackToAllocation and then not passed to sendPackOnConnection at
sidecar-handler.ts:2821, and that function has no signal parameter at all
(sidecar-handler.ts:2713-2721). sendAgentDeployToAllocation has the same shape at :3149. This
is a gap in a capability the branch adds rather than a regression, and the harm is bounded to
an orphaned pack entry and timer on a connection the Hub is abandoning. But as it stands
cancellation is honoured only before the frames go out, which is the window that was never
slow. Both new transport tests abort inside validateSidecarIdentity, so neither reaches an
ack wait.

The in-operation lease precheck has no test. I deleted the isReconciliationLeaseCurrent block
from withLeaseHeartbeat at reconciler.ts:274 and the package still ran 815 pass, 0 fail. The
fake store hardcodes it to true and nothing overrides it. That's the guard stopping a Hub
whose lease was stolen from issuing a provisioner call. Separately, runSidecarOperation has
no test file and excludedAllocationIds never runs against Postgres.

Smaller things, in case they're useful. leaseCondition now sits on the SELECT FOR UPDATE as
well as the UPDATE in bindInitialSidecar (sidecar-allocation-store.ts:480), but FOR UPDATE
freezes the row and not clock_timestamp(), so a lease that lapses in between produces a
"changed before update" throw where bindAndEnsure expects null, and a routine handoff logs as
an error. claimNextReconcilable still mints with now() at sidecar-allocation-store.ts:890
while leaseCondition and extendReconciliationLease moved to clock_timestamp(), which is
harmless at the 60s default but is the one place the two disagree. The lease-currency
predicate now has two implementations, leaseCondition and a hand-rolled copy in
updateAnchorPublicKeyUnderAllocationLock. Probe cleanup lost its only per-probe diagnostic:
the AggregateError and the message-only logging both match what main did, but the warn
deleted from initialize() was the only place the probe id and provisioner error appeared, and
this branch makes that the sole destruction path. And createHubServer returns no shutdown
handle, so the scheduler's stop() has no caller.

On docs, IMPLEMENTATION.md inserts seven sentences between "The allocation reconciler uses
database leases..." and "It persists a sidecar identity...", so "It" now reads as the probe
cleanup scan; that paragraph is fourteen sentences and has two separate disconnect sentences.
It also says a disconnect cancels unconditionally, which isn't so during provisioning.
SIDECAR_PLACEMENT.md is the provisioner-facing doc and never mentions the new abort signal or
deadline, which currently lives only in the TSDoc on contracts.ts.

One more thing worth saying. The branch also moves the periodic dispatch poll onto wake(),
where main called reconcileUntilIdle directly and bypassed the drainPromise guard. I don't
think that was a correctness bug, since claimNextPending uses FOR UPDATE SKIP LOCKED and
concurrent drains claim disjoint rows, but it's the tidier shape.

The parts I looked at hardest hold up: the runSidecarOperation timeout machinery, the split
between ensure and destroy timeouts, the AbortController lifecycle, the probe-cleanup dedup
map, releasing durability across a crash, and cross-instance claim disjointness. The store's
lease guard coverage is complete too, including beginUnrecoverableRelease, which reads as
unguarded at its call site but delegates to beginRelease.

alexanderguy (approved):
LGTM

PR #187, opened 2026-09-14
Admit inbound workflow mail by relational accept-rules

alexanderguy:
## Summary

- A workflow declares a relational `mailAccept` policy (invoker/self/tenant/correspondent toggles plus explicit principal/definition ids); the capability walk emits these as `mail.accept` approval markers into the content-hashed, operator-approved surface. Every relation it accepts is enforced, and an absent policy accepts nothing.
- The hub resolves an authenticated sender to durable admission coordinates (principal, definition, tenant) and binds a mail-triggered run's sender as its invoker.
- A run's static `mail.accept` grants materialize at launch; an admission gate at the mail-transport seam authorizes each inbound sender's coordinates against them by reusing the authz grant engine — namespace-filtered so a broad `*`/`mail.*` cannot auto-admit, deny-wins, default-deny, and fail-closed on an unresolvable sender or an unevaluable conditioned grant.
- `correspondent`: when a live run mails a party, a `mail.accept:principal:<recipient>` grant is minted on the sending run at the send instant (marker-gated, terminal-inert, idempotent), so that party's reply is admitted by the same gate.
- The admission drop log names the refused sender.

## Verification

- `make all` passes (lint, `tsc -b`, admin-ui bundle, and the unit/workflow/core suites; 0 failures).

Closes INTR-165, INTR-430, INTR-431, INTR-510, INTR-432

Delivers the inbound correspondent-accept grant of INTR-513 (bounded by terminal-run inertness); its outbound verified-signer reply addressing, and admission operability/docs (INTR-433), remain open. The `parent`/`child` relations are removed as unenforceable on the current runtime (in-process spawned children have no distinct mail identity) and are parked as a separate project.


alexanderguy:

PR #188, opened 2026-09-15
Refuse a workflow park that nothing upstream can answer

alexanderguy:
## Summary

A step that waits on a signal with no timeout, in a run nothing upstream can
answer, fails at the wait naming the gate, instead of suspending forever.

The behavior this replaces was not a working capability. A park beneath a
terminal `childWorkflow` child was already unreachable: the child's env carries
no notify sink, so the approval row never reached the tenant, nobody learned a
decision was wanted, and the parent waited on a terminal that never came. Child
runs are excluded from restart discovery, so the run could not be resumed
either. This converts that silent, unrecoverable hang into a loud failure that
names the step and the signal.

The refusal covers every untimed park, not only an `awaitSignal` an author
writes: an agent step suspending on a tool declared `approval: "ask"`, and the
input park of a step with a trigger budget, are refused the same way. A gate
carrying a `timeout` is unaffected, since its own timer resolves it in process.

## How it works

The runtime env records whether a run tree has an upstream signal resolver.
The seam that spawns a run declares it, because that is the only layer that
knows which case the run is in. The field is required rather than optional: an
absent flag would have to mean permissive, and a host that forgot it would
silently inherit the unanswerable park. Both the runtime env and the local run
surface require it, so a new call site cannot land undeclared.

Four abort bridges that lost an already-fired signal now consult its level
before subscribing to its edge, and the pattern lives in one helper. Three left
a run waiting forever; the fourth ran an action handler for a cancelled run.

## Documentation

The loop-body guidance, the authoring rules, the host contract, and the
primitive doc comments each state the boundary, so an author learns it before
writing the workflow rather than after a run goes quiet. The host contract
carries the decision procedure rather than only the field name, because the
compiler catches an omission but not a wrong value.

## Verification

`make all` exits 0: 7655 unit, 196 workflow-deploy, 756 integration, no
failures. Every commit builds independently.

Against a real hub and sidecar, both park shapes reach a terminal instead of
going quiet: an author-named gate, and the approval-shaped park where a child's
tool call suspends on the control plane. A timed gate in the same position
still runs. Each guard was verified by reverting it and watching its test fail.

Closes INTR-533


alexanderguy:

PR #189, opened 2026-09-16
Make workflow deployment and recovery more reliable

FarinM:
## Why

The deployment and recovery flow has been a bit janky around timeouts, reconnects, and partial failures. The earlier reconciliation changes improved progress across unrelated workflows, but there were still cases where timed-out work kept accumulating, a failed database lookup caused a healthy worker to be released, or a lost response made a completed deployment look unfinished.

This PR cleans up those remaining rough edges. The aim is to make retries, recovery, and shutdown more predictable: keep unfinished work bounded, preserve healthy workers when checks are inconclusive, and use durable deployment state to resolve interrupted attempts.

## What this adds

- **Bounded reconciliation work.** Allocation claims, readiness checks, lease queries, and connection writes continue counting against the concurrency limit until they actually settle, even after a timeout. Retries cannot keep accumulating unfinished database operations or reclaim an allocation whose previous work is still running.

- **Continuous lease renewal.** One heartbeat covers the entire reconciliation attempt, including database transitions between provisioning stages. Lease loss or validation errors stop the attempt and allow recovery through lease expiry without queuing another write behind stalled work.

- **Healthy-worker preservation.** Failed or unfinished identity checks leave readiness unknown and trigger a retry. They no longer count as evidence that a worker disappeared or that its running workflow needs restoration or redeployment.

- **Durable initialization tracking.** A new `initialization_lease_id` column records the deployment attempt before its frame is sent. Completion publishes the signing key and credential references and clears the marker atomically. Recovery can handle uncertain sends, roll back confirmed unsent attempts, and avoid releasing a worker whose initialization committed but whose response was lost.

- **Correct deferred-mail recovery.** Mail sent during initialization stays pending until the deployment outcome is known. Old callbacks cannot settle a newer attempt, and recovered messages retain their matching signing key. Key and mail frames are sent consecutively after asynchronous preparation completes.

- **Independent dispatch retries.** Periodic dispatch can progress independently of the enqueue notification drain. Both share a limit on unfinished work, retain timed-out deliveries until their operations settle, and check cancellation before sending late mail or signals.

- **Database statement deadlines.** Application connections default to a 60-second PostgreSQL statement timeout, configurable through `DB_STATEMENT_TIMEOUT_MS`, including time spent waiting for locks.

- **Shutdown and test-harness fixes.** Operation deadline timers no longer keep the process alive by themselves. The test-only local provisioner serializes provisioning and cleanup per allocation and drains admitted operations during shutdown.

- **Regression coverage and documentation** for stalled queries, lease expiry, ambiguous initialization outcomes, deferred-mail ordering, identity-validation failures, and database timeout behavior.

- **Retiring-allocation settlement and script timeouts.** Retiring an allocation settles the same sender state its fence settled, so a lingering sender attempt cannot act after retirement. The `bin/` database scripts also honor `DB_STATEMENT_TIMEOUT_MS`.


alexanderguy (commented):
## Review

**Agent disclosure:** this review was produced by an AI agent (Claude Code) running this repository's `pull-request-review` skill. Everything below was verified against a checkout of the branch — commands and results are cited so you can reproduce or contradict any of it. Please treat the judgement calls as a starting point for discussion rather than a verdict; I have been wrong several times in the course of preparing this and corrected myself.

---

This is careful, high-quality work. Detailed findings and inline comments below; the formal request is in a follow-up review.

The concurrency discipline is good in ways that are easy to get wrong and didn't: the refcounted exclusion map handles the delete-at-zero case correctly, admission slots release on all three exit paths without double-decrementing, the lease heartbeat is split from an independent expiry watchdog so a stalled renewal still aborts the claim, and durable writes are fenced on the exact attempt marker so a stale completion cannot clobber a newer one. All 13 commit messages pass the project's conventions. Documentation was updated alongside the code it describes.

### Verification performed

`make all` exits 0 on a clean tree with the database configured — 8,561 tests, zero failures, reproduced twice at `6e69e148`. Migration `0093`'s schema, SQL and snapshot agree, proven by regenerating with `drizzle-kit` ("No schema changes, nothing to migrate"); its validating scan measured 74ms on a 500k-row copy. No binary deltas, no out-of-scope files, `docs/API.md` is not stale.

### The one thing I would like before merge

The `!cause.frameSent &&` rollback guard is unpinned — see the inline comment. Deleting it leaves the **full** 8,561-test suite green, so nothing anywhere observes it. I checked this against the whole suite rather than a subset precisely because a subset result would not have been trustworthy.

### Two coverage gaps that are not worth holding for

`recoverSenderDeploy`'s `{failed}` arm and the identity catch in `repairUnscheduledConnections` also survive the full suite, but both have working backstops (a TTL that reaches the same terminal outcome, and two separately-tested defence layers respectively). Worth folding in if you are touching these files anyway; not worth blocking on. Inline comments on both.

### Smaller items, all new surface on this branch

`queueReconciliationStep` and the statement timeout do not exist on `main` at all, `validationFailure` is new, and `allocatedKeyRecordInFlight` changed from a `Set` to a `Map` here — so these are follow-through on this change rather than drive-by cleanup, which is why I am raising them here instead of suggesting follow-up issues. All are small. Three have inline comments; the fourth does not, because it is the *absence* of a change:

**`retireAllocation` does not settle what `fenceAllocation` settles.** The `Set`-to-`Map` change made `noteSenderDeployStarted` throw on a lingering marker where it used to be an idempotent add, and `fenceAllocation` gained a settle loop that `retireAllocation` did not. Driven directly this reproduces, but I could not find a call site that reaches it: the transition out of `allocated` bumps the generation and fences with the advanced value, settling the attempt first. Worth closing as defence-in-depth by widening the predicate to `attempt.generation <= generation`, keeping the constraint in the layer that owns fence lifecycle.

**`resolveDbConfig` cannot see `DB_STATEMENT_TIMEOUT_MS`** (`bin/lib/db-config.ts`, outside this diff so no inline comment). There was no statement timeout at all before this branch, so `bin/backfill-principal-keys`, `bin/rekey-credential-secrets` and `bin/audit-sender-keys` now run under a hard 60s ceiling this change imposed, with no override path — while their own headers tell the operator to source `.env.hub`, which is exactly where the new knob landed. `rekeyCredentialSecrets` issues one unbounded whole-table read. The docs correctly scope the variable to the hub so nothing is false, but the scripts inherited a cap they cannot adjust.

### Housekeeping

The branch is 33 commits behind `main`, and `main` has since modified files it touches, including some of its tests. It needs a rebase, and the verification that counts before merge is the rebased run rather than mine.

Separately: INTR-536 was closed on 15 September by #185. This branch carries substantial new work beyond that issue — a durable initialization column with a recovery state machine, and a new operational control — so it may want its own issue for traceability. And `af6c02cc` is 24 files and +7404/-340, an order of magnitude larger than any other commit here; its message does name each area and the column plus its consumers cannot land separately without breaking the build in between, but the deferred-mail and claims-exclusion work inside it look separable.

packages/hub-sessions/src/session-service.ts
```diff
@@ -817,6 +844,20 @@ async function emitSourceRefDeployFrame(
     };
   } catch (cause) {
     if (!isDeployFrameFailure(cause)) throw cause;
+    // Only a confirmed reservation gives us the key to restore. An ambiguous
+    // reservation response leaves its durable marker for conservative cleanup.
+    // A confirmed unsent attempt can roll back even after lease cancellation.
+    if (!cause.frameSent && previousPublicKey !== undefined) {
```
alexanderguy, line 850: Could we add a test pinning this guard? Deleting `!cause.frameSent &&` leaves the full 8,561-test suite green, so nothing currently observes it.

It matters on a **redeploy over an anchor that already carries a key**. Without the guard a *sent* frame takes the unsent-rollback path, which restores the previous key and clears the marker — and `initializationCompleted` reads marker-null plus key-non-null as *committed*. So `beginReplacement` returns null and the worker that actually received the frame is kept rather than fenced.

The part that worries me more: `onUnsentInitializationCleared` then settles the parked mail under the *previous* deployment's signing key while a possibly-live new child holds a different one. That publishes wrong key material to recipients, not just leaked capacity.

First deploys are unaffected — a null previous key makes `initializationCompleted` return false and replacement proceeds correctly. The reachable case is the redeploy, and it looks live: `workflow-allocation-service` only short-circuits when the workflow is active *and* the anchor key is non-null, so a worker that reconnected without a live child redeploys with a non-null previous key.

One test asserting that a `frameSent: true` failure with a confirmed reservation leaves the marker in place and does not settle parked mail with the restored key would close it.

packages/hub-sessions/src/session-service.ts
```diff
@@ -967,6 +1008,36 @@ export async function deployCodeSourcedWorkflow(
...
+  const anchor = await args.db.query.workflowRun.findFirst({
+    where: eq(workflowRunTable.id, allocation.anchorRunId),
+    columns: { publicKey: true },
+  });
+  reconciliation.signal.throwIfAborted();
+  args.sidecarRouter.noteSenderDeploySettled(
+    { allocationId: allocation.id, generation: allocation.generation },
+    anchor !== undefined && anchor.publicKey !== null
+      ? { recorded: anchor.publicKey }
+      : { failed: "Previous deployment initialization did not complete" },
```
alexanderguy, line 1037: Minor: this arm's outcome is never asserted — replacing it with a `{recorded}` settlement leaves the full suite green.

Not worth blocking on, since the deferred-mail TTL reaches the same terminal outcome (claims the entry, emits `mail.outbound.undelivered`, logs a warning), so the worst case is delay rather than lost mail. Worth noting the function itself is well covered — deleting its early-return guard does fail two existing tests. It is specifically this arm that is unpinned.

packages/hub-sessions/src/sidecar-allocation/reconciler.ts
```diff
@@ -818,6 +931,37 @@ export function createSidecarAllocationReconciler({
...
+            await apply();
+          }),
+        active.controller.signal,
+      );
+    } catch (error) {
+      if (error instanceof SidecarOperationTimeoutError) {
+        // Stop this lease without queuing another write behind the same stalled
+        // event. Keep the actual operation queued until it settles, preserving
+        // ordering with reconnects and excluding this allocation from new claims.
+        const cancelled = new ReconciliationLeaseLostError(target.allocationId);
```
alexanderguy: Should this carry the timeout as its cause? As written the `SidecarOperationTimeoutError` is dropped entirely, and because `reconcileClaim` branches on `cause === undefined` it then logs at **info** that the lease "is no longer current" — which is not what happened when a local per-allocation event queue exceeded its deadline. The only diagnostic for that path is gone.

`new ReconciliationLeaseLostError(target.allocationId, error)` would surface it through the existing warn branch, and matches the `{ cause }` convention in CONVENTIONS.md.

packages/hub-sessions/src/sidecar-allocation/reconciler.ts
```diff
@@ -876,7 +1020,14 @@ export function createSidecarAllocationReconciler({
         allocationId: allocation.id,
         generation: allocation.generation,
       };
-      if (await router.isAllocatedSidecarReady(target)) continue;
+      try {
+        if (await router.isAllocatedSidecarReady(target)) continue;
+      } catch (error) {
+        // Unknown readiness is not absence. Leave the allocation for the next
+        // repair sweep instead of scheduling a reconnect the worker may hold.
+        if (error instanceof SidecarIdentityValidationError) continue;
```
alexanderguy, line 1031: Minor: this catch is untested — deleting the whole `try`/`catch` leaves all 113 reconciler tests green, and the only two tests touching `repairUnscheduledConnections` use `ready: false` / `ready: true` stubs that never throw.

I am not suggesting this blocks: it is the third layer of a defence whose other two layers *are* covered (`waitUntilReady` rethrows `SidecarIdentityValidationError` instead of calling `replaceAfterFailure`, and `waitForAllocatedSidecar` reports an inconclusive validation rather than a missed deadline), so removing it cannot release a healthy worker. But since this is the headline behaviour of "Preserve healthy workers when identity validation is inconclusive", it seems worth an assertion — `deps()` already supports `readyError`, so it should be a few lines next to the existing repair tests.

packages/hub-sessions/src/ws/sidecar-handler.ts
```diff
@@ -1382,15 +1423,49 @@ export function createSidecarRouter(
...
+      // revalidates, and at expiry the wait reports the failure rather than a
+      // missed deadline. Registration itself was already gated, so this must
+      // not fail the connection that just registered.
+      const validationFailure = new SidecarIdentityValidationError(
+        allocationId,
+        current.identity.generation,
+        cause,
+      );
+      for (const waiter of matchingWaiters) {
+        waiter.validationFailure = validationFailure;
```
alexanderguy, line 1451: This field is assigned here and never cleared anywhere, so the first bad observation outranks any later, better-informed one — the timeout callback reads it unconditionally.

Concretely: one transient lookup failure presets it, the worker then disconnects for good, and at expiry the caller gets a `SidecarIdentityValidationError` (which the reconciler deliberately treats as inconclusive and retries) instead of the plain connection timeout it should get. That is the exact distinction this commit introduces, inverted by a stale field.

Bounded to one extra reconciliation cycle, since the next pass recomputes `remaining` from the durable `connectDeadline` and takes a fresh reading — so this is a precision issue, not a stall. Clearing it whenever a validation completes without throwing (on both the `true` and `false` outcomes) would keep it holding current evidence.

packages/hub-sessions/src/ws/sidecar-handler.ts
```diff
@@ -2660,11 +2817,18 @@ export function createSidecarRouter(
             allocationWaiters.delete(target.allocationId);
           }
           reject(
-            new Error(
-              `Timed out waiting for allocated sidecar ${target.allocationId} generation ${String(target.generation)}`,
-            ),
+            waiter.validationFailure ??
```
alexanderguy, line 2838: This is where the stale `validationFailure` wins — it is preferred over the freshly-computed local value, so an early transient failure masks later definitive evidence. See the note at the assignment site.

alexanderguy (changes requested):
## Requesting one change

**Agent disclosure:** posted by an AI agent (Claude Code) running this repository's `pull-request-review` skill, following up on the detailed review above to make the state match the ask.

The branch is in good shape and I have no correctness objection to it. There is one thing I would like before merge:

**Add a test pinning the `!cause.frameSent &&` rollback guard** in `emitSourceRefDeployFrame` (inline comment on `packages/hub-sessions/src/session-service.ts`).

Deleting that condition leaves the **full** 8,561-test suite green, so nothing currently observes it. On a redeploy over an anchor that already carries a key, losing it means a *sent* frame takes the unsent-rollback path, the durable state then reads as committed, and the parked mail is settled under the *previous* deployment's signing key while a possibly-live new child holds a different one. Wrong key material published to recipients is a bad failure to have unguarded by a test, even though the code is correct as written today.

One test asserting that a `frameSent: true` failure with a confirmed reservation leaves the marker in place and does not settle parked mail with the restored key would close it.

Everything else in the review above is a suggestion, not a condition:

- `recoverSenderDeploy`'s `{failed}` arm and the `repairUnscheduledConnections` identity catch are also unpinned, but both have working backstops. Worth folding in while you are in these files; I am not gating on them.
- The discarded timeout cause in `queueReconciliationStep`, the uncleared `validationFailure`, the `retireAllocation` asymmetry, and the `resolveDbConfig` statement-timeout gap are all small and all new surface on this branch, which is why I raised them here rather than as follow-up issues. Your call on each.

Happy to re-review promptly, and happy to be talked out of the above if you see coverage I missed.

packages/db/README.md
```diff
@@ -37,6 +37,17 @@ typed values; downstream code uses those values without
 re-casting. See `CONVENTIONS.md` for the project-wide rule against
 cast-at-callsite in DB consumers.
 
+Connections created by `createDB` set PostgreSQL's `statement_timeout` to
+60,000 milliseconds. Override it with `statementTimeoutMs` in `DBConfig`
+(a positive integer up to 2,147,483,647 milliseconds). The Hub reads this
+override from `DB_STATEMENT_TIMEOUT_MS`.
```
alexanderguy: Could this line pick up the widened scope from `1f19bed36`? That commit taught `resolveDbConfig` to read `DB_STATEMENT_TIMEOUT_MS` and updated DEV.md's scope column to "hub (`apps/hub`), db scripts (`bin/`)", but this sentence still names only the Hub — so the package README now understates who honours the override, which is the one place a reader of `@intx/db` would look.

```suggestion
(a positive integer up to 2,147,483,647 milliseconds). The Hub and the
`bin/` database scripts read this override from `DB_STATEMENT_TIMEOUT_MS`.
```

alexanderguy (approved):
## Re-review: approving

**Agent disclosure:** produced by an AI agent (Claude Code) running this repository's `pull-request-review` skill, following up on my review of 18 September. Every affirmative claim below names the check behind it, so you can reproduce or contradict any of it.

The one condition from that review is met, and all five suggestions were taken as well. I have no remaining objection to merging.

### The gating request is closed

`tests/db/anchor-before-frame.test.ts:546` pins the `!cause.frameSent &&` guard. I checked it by mutation rather than by reading it: with `session-service.ts:850` changed to `if (previousPublicKey !== undefined)`, that file goes from 12 pass / 0 fail to 11 pass / 1 fail, and the one failure is that test, on the `initializationLeaseId` assertion. The test drives the case I described — a redeploy over an anchor that already carries a key, a `frameSent: true` failure after a confirmed reservation — and asserts the marker survives, the previous key is not restored, and the parked mail is neither dropped nor delivered.

### The five suggestions

I mutation-tested each new test the same way. In every row the named test is the one that fails, and the only one:

| Prior comment | Where it landed | Mutation that proves the pin |
| --- | --- | --- |
| `recoverSenderDeploy`'s `{failed}` arm unpinned | `session-service.test.ts:1691` | arm replaced with a `{recorded}` settlement |
| `repairUnscheduledConnections` identity catch untested | `reconciler.test.ts`, "leaves an allocation with inconclusive readiness for the next repair sweep" | catch falls through instead of `continue` |
| `validationFailure` never cleared | `sidecar-handler.ts:1492-1496`, test in `sidecar-handler.lifecycle.test.ts` | clearing loop neutered |
| `retireAllocation` does not settle what `fenceAllocation` settles | `f4736497b`, `sidecar-handler.ts:2699-2710`, test in `sidecar-handler.mail.test.ts` | settle loop deleted |
| timeout cause discarded in `queueReconciliationStep` | `reconciler.ts:957-960` | read the path instead; see below |

On the three that are more than a test:

**`retireAllocation`.** The `attempt.generation <= target.generation` predicate is the right widening: the function returns early unless the fence still equals `target.generation`, so `<=` covers the current attempt plus any stale earlier one, while `fenceAllocation`'s `<` correctly leaves the incoming generation's attempt live. I read all five call sites (`reconciler.ts:604`, `:762`, `:813`, `:876`, and `workflow-allocation-service.ts:349`) — each runs after a durable `failed`/`released` transition or a completed probe cleanup, so none can reach it with an initialization that might still legitimately complete. The snapshot-then-settle shape is also safe against the inner `noteSenderDeploySettled` mutating the same map.

**The timeout cause.** Threaded now, and I traced it to the consumer: `reconciler.ts:1176-1182` branches on `error.cause`, so a `SidecarOperationTimeoutError` reaches the `warn` branch carrying its message instead of logging "is no longer current" at info.

**`resolveDbConfig`.** Plumbed end to end — `bin/lib/db-config.ts` → `createDB` → `connection.ts:20` reads `config.statementTimeoutMs` — and the `set -a; . .env; . .env.hub` recipe in the script headers now actually carries the override, since `.env.hub.example:4` is where the knob is documented. Three boundary tests cover set, unset/empty, and non-positive-integer.

### One small thing, not a condition

`packages/db/README.md:43` still says the Hub reads `DB_STATEMENT_TIMEOUT_MS`. `1f19bed36` widened DEV.md's scope column to "hub (`apps/hub`), db scripts (`bin/`)" but left this line naming only the Hub, so the README now understates who reads the variable. Inline comment with a suggested one-liner.

Separately, the PR description predates the last two commits, so it does not mention the retiring-allocation settle or the scripts' access to the timeout knob. Worth a refresh before merge if the description is going to serve as the squash message.

### Housekeeping from last time

The rebase happened: the merge base moved from `c611c0c` to `bca283b`, and the branch is 13 behind `origin/main` rather than 33. `git merge-tree --write-tree origin/main HEAD` exits 0 with no conflict output, and `docs/IMPLEMENTATION.md` is the only file that both main's drift and this branch touch, so there is nothing to untangle.

One rebase resolution deserves calling out because it looks like a lost upstream fix and is not. `0dd83f402` deletes "parks the allocation when lease renewal throws transiently" — the test main had just stabilized with the `leaseDurationMs: 300` ratio comment. The deletion is correct: this branch changes that behaviour, so a renewal failure no longer parks, and the replacement "a renewal failure finishes without parking while its read retains capacity" asserts the new outcome by racing a `parkEntered` resolver instead of waiting out a duration — the better shape under "Synchronizing on State, Not Time". Main's timing rationale survives on the test where it still applies.

`2573a664a` is still the outlier at 24 files, +7603/-340, though 4,407 of those insertions are the generated `0093_snapshot.json`, so roughly 3,200 lines are hand-written. I suggested splitting it last time and you kept it whole; I am not re-raising it.

### Verification performed

`make all` exits 0 at `1f19bed36` on a clean tree with the database configured: 8,820 tests pass, 0 fail, 2 skipped across the three passes (7,841 unit / 195 workflow-deploy / 784 core). Lint is clean, including the `docs/API.md` freshness check, `check-test-enumeration` (595 test files covered), and `check-ci-test-jobs`. The six `error:`-shaped lines in the log are expected output from negative-path tests with deliberately corrupt fixtures, not failures.

Commit hygiene: `git log --format='%s' | awk 'length > 72'` and the same over `%b` both return nothing; no subject carries a prefix, trailing punctuation, or a filename; no body references a tracker, another commit in the series, or review discussion. History is linear, single-author, with no `Co-Authored-By` trailers. `git diff origin/main...HEAD --stat` shows 38 files, no `Bin` markers, nothing outside the branch's scope, and no `package.json` change that would need a lockfile alongside it.

### One limit on the above

The `validationFailure` clearing closes the sequential case I described. It does not close an overlapping one: `notifyAllocationWaiters` can run concurrently (`sidecar-handler.ts:2890` fires one per waiter, the register path awaits another at `:1672`), and each invocation issues its own readiness validation, so a clean reading that settles first can still be overwritten by an older failing one settling after. That is the same bounded, self-correcting severity as the original nit — the next pass recomputes from the durable `connectDeadline` and takes a fresh reading — so I am not asking for a change. I am recording it so the field is not read as unconditionally holding current evidence.

Nice work on this one. Thanks for taking the whole list.

PR #190, opened 2026-09-17
Let inference call tools wherever it runs

alexanderguy:
## What this branch does

Inference can call tools anywhere it runs. Before this branch, a step nested
inside a `loop` body, an `onTrigger` section body, or a `childWorkflow` body
could not invoke a tool in a deployed workflow — the call was refused before the
tool ran.

The refusal was silent, which is why it survived. `createToolRunner.run`
converts a handler throw into a well-formed error `ToolResult`, so the authorize
failure became an ordinary tool result, the agent answered it in text, and the
run reached `RunCompleted`. Nothing failed; the work simply never happened.

## Root cause

Three separate step-walks existed and disagreed. `normalize` built `stepOrder`
from top-level steps only, while the capability walk and the source-pinning pass
each recursed into nested bodies. The credentials snapshot was keyed off the
non-recursive one, so a nested step had no entry, and the authorize closure —
which throws on a missing entry rather than defaulting — refused every call.

The fix promotes one canonical walk and wires all three snapshot producers to
it, so the set of steps that can run and the set of steps that have grants are
derived from the same traversal. The walk dispatches nested bodies through an
exhaustive switch with a `never` assertion, so a newly-added container kind
fails at compile time instead of silently reading as a leaf.

## Also in this branch

- The deploy gate refuses a deployment whose executable closure reaches a step
  the grant walk left no record for, and gates declared grant requirements
  against operator approval.
- The gate refuses reserved-but-unimplemented `schedule` triggers, which
  previously deployed and then never fired — no error, no log, no failed run.
- The local runtime requires an authorize decision and honors it. It previously
  defaulted to permitting everything, so a workflow could pass locally and fail
  on deploy; local runs were the safety net and the net measured nothing.
- A loop publishes its converging iteration as `final` alongside `carry`.
- The approval policy is a compiler-checked discriminated union. It was
  distinguished by testing whether one optional key was present, which would
  have resolved every operator approval to the approve-everything arm had the
  approved-surface type ever grown that key.
- A runnable workflow example and an authoring guide, plus corrections to
  documentation that described mechanisms which do not exist.

## Why the branch is clean

- `make all` exits 0: 8,580 tests pass, 0 fail.
- All 33 commits build in isolation, so none depends on a later one to compile.
  Each was checked out individually and passes `make lint && make build`.
- The tool-invoke tests assert the tool's returned **text**, not the presence of
  a `tool_result` block. A presence-only assertion passes on a fully broken
  system, because a refusal still produces a well-formed error result.
- Commit-message audits over the range return empty for over-72 subjects and
  body lines, subject prefixes, filenames in subjects, trailing punctuation, and
  tracker or cross-commit references.
- `git diff origin/main...HEAD --stat` shows no binary markers and no files
  outside scope. The three infrastructure files that change — the root
  `tsconfig.json`, the `Makefile`, and `bun.lock` — register the new example and
  test projects; the enumeration check fails without them.
- Generated API docs are current: the gate's `gen-api-docs --check` step reports
  `docs/API.md is up to date`.

## Scope note for the reviewer

Any change to a type that `parse-row.ts` asserts against a persisted column must
be diffed against `origin/main`, not merely reviewed on the branch. This branch
came close to shipping a change that renamed a persisted field and converted it
from a string list to a struct; both halves would have rejected every stored row,
and the full gate was green throughout, because no fixture held a row written by
an older build. `FrozenApprovalBundle` now has a test that does.

## Review rounds

Four independent review passes, the last with four agents on separate lenses
(authorization, persisted shape, runtime execution, test quality). No fail-open
authorization path; the one durable type change is a strict widening verified at
the real row boundary; the wire-hash preimage is provably unmoved; none of the
five tool-invoke round-trips is vacuous.

Three defects that review found and this branch fixes:

- A nested body step id was not bound to the key it is recorded under, so the
  grants domain and the runtime authorize key could name different steps. That
  is the exact divergence this branch exists to close, found inside the
  branch's own new code. The fix is mutation-tested: the new case fails without
  it.
- A gate test looped over an always-empty list and therefore executed no
  assertion. It would have passed against a gate that approved two forged
  requirements and merely dropped them from the snapshot.
- `parseWorkflowRunLaunchSpecRow` had no test anywhere. It now has one driven by
  a populated row of the older approval shape; narrowing the field back fails it.

Review also closed two gaps this branch's own change opened: the loop output
record is rebuilt from a different source on a replayed settle than on a driven
one, and nothing held the two to agreement across a mid-flight resume — there
are now cases at nine crash boundaries. And the record was documented as
persisted inline when it embeds the last iteration's whole output tree and can
spill to a blob, which a loop container's output never previously reached.

One known limitation is documented rather than fixed: an `onTrigger` section
body can silently skip a later run of the same deployment, because its run id
carries no parent-run prefix. Pre-existing and untouched here, but this branch
makes those bodies tool-bearing, so what a skipped body skips is now real work.
Changing that derivation moves a durable key and needs its own migration, so it
is tracked as INTR-552 rather than ridden along on this branch.

Closes INTR-545
Closes INTR-548

Follow-up issues filed from this work: INTR-549, INTR-550, INTR-551.


alexanderguy (commented):
## Self-review

Reviewed the full `origin/main..HEAD` range: 33 commits, 167 files, across four passes —
the last with separate lenses on authorization, persisted shape, runtime execution, and
test quality.

Verified:

- `make all` exits 0 at 8,580 passing, 0 failing
- Every commit builds in isolation, so none depends on a later one to compile
- The deployed tool-invoke tests assert the tool's returned text, not the presence of a
  `tool_result` — a presence-only assertion passes on a broken system here
- Commit-message, scope, and binary audits over the range are clean

Not verified: no manual run against a live hub and sidecar. Three open questions are
recorded in INTR-550, INTR-551 and INTR-552 rather than left implicit.

One known limitation ships documented rather than fixed: an `onTrigger` section body can
silently skip a later run of the same deployment. Pre-existing, unchanged here, tracked
as INTR-552, and written into the authoring guide.

**Clean and ready to merge.**

PR #191, opened 2026-09-18
Synchronize tests on state instead of elapsed time

alexanderguy:
## Summary

- Tests wait on a signal that proves the awaited state was reached, rather
  than on a duration chosen to be long enough. `CONVENTIONS.md` states the
  rule and a second one requiring a comment's guarantee to be true by
  construction; every fix here cites them.
- Three production timer seams are injectable, so a test drives a deadline
  instead of outwaiting it: the session layer's pending tracker and sidecar
  handler, and the workflow host's queued scheduler. Each returns a
  canceller rather than a handle, and each defaults to the global timer, so
  production behaviour is unchanged. The tracker identifies the entry a
  timeout belongs to by a per-registration token, because keys are reused
  and a stale fire that matched on key occupancy alone would drop a live
  successor and re-run a settled entry's cleanup.
- The waiting helpers the tests had copied — roughly thirty poll helpers
  under twenty-one names, and the in-memory stream doubles in eighteen
  copies — are shared, level-triggered, and carry no deadline. The LogTape
  capture harness is shared on the same terms. Each has tests written by
  mutating the helper until they failed.
- Wait budgets in the integration lane belong to the runner. The harness
  tracks its own in-flight waits instead, so a wedged test reports which
  wait was outstanding and the sidecar output beside it, and teardown stops
  those waits rather than letting one poll on against a dismantled fixture.
  `retrying` passes a torn-down check into its body, so a caller polling on
  its own is reachable by that stop rather than running past it into a
  dismantled fixture and reporting a fault of the dismantling.
- The LSP client awaits its protocol notification writes, so a failed write
  surfaces on the operation that caused it rather than as an unhandled
  rejection charged to an unrelated test. A write its bound abandons is
  neither delivered nor failed and is not cancellable, so it poisons that
  document rather than letting a later call reuse a version the server may
  already hold and report diagnostics for stale text as current.
- A respawn bail reports itself. The backoff announced the respawn it was
  about to wait out, and the woken coroutine then bailed silently when it
  found the cohort superseded.
- Several assertions could not fail and now can: an ordering proved by
  comparing two `Date.now()` readings with `>=`, a reconnect test re-reading
  an earlier run's terminal event, a run-tolerates-abort test whose only
  check was that a timer lost, a poll that fell through silently so a gate
  that never parked passed, and a two-turn mail test that could not tell a
  mid-run wake from a `mail_wait` expiry.
- A two-sidecar deployment names the transport it binds to. Inferring it
  from which sidecar was connected raced a reconnect, which co-located two
  deployments that must be split and made a send local that must be remote.

## Behaviour change to note

`SIDECAR_RECONNECT_DELAY_MS` is validated with `Number` rather than
`Number.parseInt`, so it rejects the whole string instead of reading a
leading numeric prefix. `3000ms` and `5abc` previously arrived as `3000` and
`5`; `1e5` arrived as `1`, turning an intended 100-second backoff into a 1ms
reconnect loop against the hub. Those now fail the boot, which is what the
reader's comment already claimed. Production sets this variable nowhere in
the tree; every setter is a test fixture passing an integer.

## Verification

`make all` passes. The unit pass is green at full worker parallelism — the
condition that surfaces this class — across repeated runs.

Three of the restored assertions are verified by differential mutation
rather than by inspection: with a one-line mutation applied to
`supervisor.ts`, `origin/main`'s tests fail and this branch's now fail too,
where before this branch's passed. The mutations are the durable-consume
gate in the signal-deliver dispatch, the `selfTerminated` phase allowlist,
and the generation arm of the post-backoff respawn guard.

Every commit is independently lint-clean: the pre-commit hook checks each
against a staging tree, including formatting, lint, API-doc freshness,
dependency and export shape, and test enumeration. Two windows where the
branch did not type-check are closed — the shared testing entry point now
lands before its first consumer, and a call that outlived its helper's
rename is fixed at the commit that renamed it — but the branch has not been
swept commit-by-commit with `make build`, so treat per-commit
type-checkability as fixed-where-found rather than verified throughout.

## Known limitation

An LSP document poisoned by an abandoned write stays poisoned for the
client's lifetime. Recovery would need a `didClose` ordered behind the
stalled write, which the JSON-RPC writer's semaphore would give, but that is
a larger change than this branch should carry. Failing loudly on that path
is a strict improvement on serving stale diagnostics as current.


alexanderguy:
Self-review (`code-review` skill) found four issues, all now fixed on the branch.

A commit bundled comment-only Makefile changes its message did not describe; it is split in two. A condition helper carried an unused parameter and an unused second export, both promised by its own comment and reached for by nothing; both are gone. Nine new helper modules had no tests, which the existing `testing` precedent in this repo pairs with each helper; they have them now.

Writing those tests corrected a claim in the helpers themselves. Six comments said a waiter is armed before the state is read so a report landing between the two is not missed. Mutation showed that ordering is not what makes them safe — the registering executor runs synchronously, so nothing can land between those two lines in either order. What makes them safe is re-checking on every pass, which is also what lets them be satisfied by state that arrived before the wait did. The comments say that instead, and the tests pin the property that survives mutation.

alexanderguy:

Reviewed `origin/main...HEAD`: 15 commits, 196 files. Clean, with the
exceptions below, each of which states its reason at the site.

**Durations kept because no signal exists.** All are negative assertions —
they prove something has *not* happened, and a predicate for that holds the
instant it is first read. Each asserts no upper bound, so overshooting under
load can only under-detect, never fail a correct run.

- `drain-roundtrip.test.ts`, `single-step-mail-wait-wake.test.ts`,
  `inbound-mail-enforcement-real-route.test.ts`, `credential-routes.test.ts`,
  `spawn-replay-fifo.test.ts`

**Durations that are the subject, not a wait.** `cache.test.ts` (11) spaces
writes for filesystem atime granularity; `step-invoker.test.ts` is a
double's own simulated work; `reactor.test.ts` drives a real production
duration through an injected seam.

**Two ordering bets I could not remove.** `single-step-mail-two-turn-thread`
and `single-step-mail-flag-expunge` arm a watch inside the workflow-child
process, which reports none of it to the harness. A lost race degrades what
they exercise rather than failing them, because the mail-wait handler
searches the mailbox before installing the watch.

**Poll intervals inside unbounded loops.** Ten sites tick while their loop
carries no deadline of its own, so the interval decides nothing and the
lane budget is the failsafe.

**Bounds kept deliberately.** `lib/http.ts` and the two Playwright harness
files bound processes the suite does not own. `latency-gate.bench.ts` is a
benchmark whose subject is duration and which no `make` target runs.
`source-tamper-reject.e2e.test.ts` uses expiry as its pass condition.
`crash-respawn-fifo.test.ts` uses it as retry cadence.
`mail-federation-run-completes.test.ts` uses it to discriminate failure
classes and to emit the diagnostic that renders both sidecars.

PR #192, opened 2026-09-18
Let agents attach files to mail and read them back

TheGreatAxios:
## Summary

- `mail_send` and `mail_reply` accept attachments on conversation mail: plain text for text-like types, base64 otherwise, validated with the same policy the hub route applies.
- The in-memory transport's send now carries attachments into the signed content. It dropped them silently before, and refuses them on structured messages.
- `mail_read` lists received attachments with a part path; fetching that path returns valid UTF-8 text as text and any other bytes as base64.

The wire format is unchanged: attachments are base64 MIME parts exactly as on `main`.

## Changes

- `packages/types`: attachment validation moved here from `hub-api` so the tools and the hub route share one policy; it also accepts already-decoded bytes. New `isTextLikeMimeType` predicate.
- `packages/mail-memory`: `send` passes attachments through and rejects them with a structured payload.
- `packages/tools-mail`: tool schemas, handlers, README, and a round-trip test against a real signing transport (adds `@intx/crypto` and `@intx/mail-memory` as devDependencies).
- `docs/MESSAGE.md`: tool parameters and the `mail.read` attachment contract.

## Testing

`make all`: lint, type-check, and the admin bundle pass; `test-workflow` and `test-core` pass. Unit tests are 7649 pass, 1 fail. The failure is the `storage-isogit` exec-bit test, which fails identically on `main` in my environment (umask) and is unrelated to this change.

## Known limitation

The `part` path `mail_read` advertises assumes the layout our own writer emits (text first, then attachments). Mail from another producer with extra inline parts would be numbered incorrectly, and the agent would fetch the wrong part without an error. Every current producer is our writer. A proper fix reports real MIME positions from message parsing and is proposed as a follow-up.

Closes INTR-559


TheGreatAxios:
## Self-Review Summary

Reviewed against `main` with the `code-review` skill, then independently by a critique agent (which exercised the real signing transport with odd inputs) and a CTO-profile reviewer, and audited again after the fixes. An earlier revision also chose 7bit / quoted-printable / base64 per attachment on the wire; that was removed as out of scope for INTR-559, so the wire format here is identical to `main`.

### Files Reviewed

- `packages/types/src/attachment-validation.ts`: moved from `hub-api` unchanged apart from imports, then extended to accept decoded bytes. The hub caller changes by one import.
- `packages/types/src/attachments.ts`: new `isTextLikeMimeType`, limited to `text/*`, JSON, and `+json`.
- `packages/mail-memory/src/send.ts`: attachments reach the signed content; structured mail with attachments is refused. The transport owns that rule since it is a wire-format fact; the tool surfaces it as `send_failed`.
- `packages/tools-mail/src/handlers.ts`, `definitions.ts`: arktype-validated input, shared validator, typed response literals, strict UTF-8 on read.
- Tests: mock-transport handler tests, a recipient-side test in `mail-memory`, and `attachment-roundtrip.test.ts`, which drives the tools through a real signing transport and checks the advertised part paths, empty attachments, non-UTF-8 text, and the structured-mail refusal.
- `docs/MESSAGE.md`, `packages/tools-mail/README.md`: match the code.

### Issues Found and Resolved

- Attachments from the tools never reached the wire because send built the conversation part from text alone; mock-only tests hid it. Fixed, with a test that asserts at the recipient.
- Reading a part decoded every byte sequence as UTF-8, mangling binary and non-UTF-8 text. Text now comes back as text only when it is valid UTF-8, otherwise base64; a leading BOM is preserved.
- The tools duplicated the hub's validation. Now one validator in `@intx/types`.
- An explicit `utf-8` encoding on a binary content type would deliver a corrupt file silently. Now refused with `invalid_encoding`.
- The original history had a tests-only commit that failed the build. Now four commits that each stand alone.
- Schema block duplicated between `mail_send` and `mail_reply`, `encoding` lacked an enum, and the read handler used untyped records. All tidied.

### Not Fixed

- The part-path numbering limitation described in the PR body. Proposed as a follow-up issue rather than a change here.
- A lone UTF-16 surrogate in text content is replaced on encode. It requires malformed model output and was left alone.
- `make all` has one failing unit test, `storage-isogit` exec-bit, which fails identically on `main` in the same environment.

packages/tools-mail/src/handlers.ts
```diff
@@ -327,12 +459,18 @@ export function makeMailReadHandler(transport: MessageTransport): ToolHandler {
...
+    const text = isTextLikeMimeType(part.contentType)
+      ? decodeStrictUTF8(part.content)
+      : undefined;
     return {
       callId: call.id,
       content: {
         contentType: part.contentType,
-        encoding: part.encoding,
-        content: new TextDecoder().decode(part.content),
+        encoding: text === undefined ? "base64" : "utf-8",
```
alexanderguy: This asserts an encoding label rather than reporting the part's own, but `fetchPart` only undoes base64 — every other transfer encoding passes through untouched.

So a `text/plain` attachment sent `Content-Transfer-Encoding: quoted-printable` with body `caf=E9` comes back as `{ encoding: "utf-8", content: "caf=E9" }`. `decodeStrictUTF8` succeeds because the raw bytes are valid ASCII, so nothing errors — the agent gets mojibake presented as decoded text. Verified end to end: the listing advertises `size: 4` while the fetched part is 6 bytes, because `extractAttachments` *does* decode quoted-printable and `fetchPart` doesn't. The same response disagrees with itself.

On `main` this line was `encoding: part.encoding`, so a caller could at least tell. And `docs/MESSAGE.md` now states the transfer encoding is "already undone", which is false for this input. Same shape for 8bit, binary, and any unrecognised encoding whenever the bytes happen to be valid UTF-8.

Either fail loudly when `part.encoding` names something `fetchPart` didn't undo, or route `fetchPart` through the same decode helper `extractAttachments` uses so the documented contract becomes true.

packages/tools-mail/src/handlers.ts
```diff
@@ -70,6 +91,88 @@ const FlagArgs = type({
...
+function attachmentsField(message: InboundMessage) {
+  if (message.attachments === undefined || message.attachments.length === 0) {
+    return {};
+  }
+  return {
+    attachments: message.attachments.map((att, index) => ({
+      name: att.name,
+      contentType: att.contentType,
+      size: att.data.length,
+      part: `1.${index + 2}`,
```
alexanderguy: The part path is derived from the position in `message.attachments`, but that array comes from a walk that skips non-attachment sub-parts, so the two indices only agree for mail our own writer produced.

For an inner multipart of `[text/plain body, text/html alternate, image/png]`, `isAttachmentPart` returns false for the inline HTML, so `extractAttachments` skips it and the png lands at `attachments[0]` — advertised as `1.2`. The real `1.2` is the HTML body. Reproduced: asking for the advertised part returns the HTML, `fetchPart` succeeds, and neither `invalid_part` nor anything else fires. The png is at `1.3`.

The docstring above is honest that other layouts would be numbered incorrectly, and the PR body calls it a known limitation on the grounds that every current producer is our writer — but the code doesn't enforce that, and `HubTransport`'s own docs name an SMTP/IMAP relay as an anticipated backend.

Security impact is limited and I don't want to inflate it: the metadata and the mis-numbered part come from the same sender, and the fetched part reports its own true content type. The correctness impact isn't limited — this hands an agent the wrong bytes in exactly the path the feature exists for. Cheapest fix is to compare the fetched part's length and content type against the advertised `size`/`contentType` and return `invalid_part` on mismatch; the thorough one derives the path from `fetchStructure`.

packages/tools-mail/src/handlers.ts
```diff
@@ -327,12 +459,18 @@ export function makeMailReadHandler(transport: MessageTransport): ToolHandler {
       );
     }
 
+    // Text comes back as text; anything else as base64, since decoding
+    // arbitrary bytes as UTF-8 would corrupt them. Mirrors the `content` /
+    // `encoding` pair the send tools accept.
+    const text = isTextLikeMimeType(part.contentType)
```
alexanderguy: Content type doesn't round-trip between `mail_read` and `mail_send`, and this line is where the two parsers diverge.

`isTextLikeMimeType` splits on `;` before matching, so it handles parameters. The allowlist `mail_send` uses does an exact key lookup and doesn't. A part read back as `"text/plain; charset=utf-8"` — which is what this response returns verbatim — is rejected `disallowed_mime_type` when fed straight back as an attachment `contentType`. Verified.

The two `mail_read` responses also disagree about the same attachment: `attachmentsField` reports the normalised `"text/plain"` while the part fetch reports the parameterised form. A model forwarding an attachment it just fetched will fail.

packages/tools-mail/src/handlers.ts
```diff
@@ -70,6 +91,88 @@ const FlagArgs = type({
...
+  // mislabelled; sending it would deliver a corrupt file with no error.
+  const mislabelled = inputs.findIndex(
+    (input) =>
+      input.encoding === "utf-8" && !isTextLikeMimeType(input.contentType),
+  );
+  if (mislabelled !== -1) {
+    return {
+      ok: false as const,
+      error: {
+        code: "invalid_encoding",
```
alexanderguy: `invalid_encoding` is documented in `docs/MESSAGE.md` but isn't a member of the `AttachmentError` union in `packages/types/src/sessions.ts`, so a caller switching exhaustively over that union never sees it.

Smaller than it first looks: `errorResult` keeps only `{error, code}`, so the structured fields every other attachment error carries are already dropped on the tool path too. The residual point is just that the docs name six codes and only five exist as a type.

packages/tools-mail/src/handlers.ts
```diff
@@ -29,20 +36,34 @@ export type ToolHandler = (
 // Argument schemas
 // ---------------------------------------------------------------------------
 
+// A tool-facing attachment: plain text unless `encoding` says otherwise. The
+// model never has to hand-encode base64 for a text file -- `content` is the
+// text itself, and `encoding` only exists to opt into base64 (or to force it
+// for a text-like type the caller already has base64-encoded).
+const AttachmentInput = type({
```
alexanderguy: Nit: this is the second `AttachmentInput` in the tree. The `@intx/types` barrel now publishes a different shape under the same name — `{mimeType, data, name?}` versus this `{name, contentType, content, encoding?}`. They never meet in one module so nothing breaks, but a reader following the import lands on the wrong shape. `AttachmentToolInput` would remove the trap.

packages/types/src/attachment-validation.ts
```diff
@@ -45,7 +46,8 @@ export type AttachmentValidationResult =
   | { ok: true; attachments: MessageAttachment[] }
   | { ok: false; error: AttachmentError };
 
-function decode(data: string): Uint8Array | null {
+function decode(data: string | Uint8Array): Uint8Array | null {
```
alexanderguy: While this is the boundary being widened: `validateAttachments` below calls `decode` on each entry in full *before* comparing its length to the limit, and keeps decoding after a rejection rather than returning early. There's also no cap on the array length.

Bounded in practice — the HTTP caller sits behind a 44 MB `bodyLimit` and the tool caller is bounded by model output — so the realistic peak is tens of megabytes, not the gigabytes a naive reading suggests. Worth a cheap guard anyway: rejecting on the base64 string length before decoding is sound, since decoded bytes never exceed three quarters of it.

packages/types/src/attachment-validation.ts
```diff
@@ -1,19 +1,20 @@
...
-import type { MessageAttachment } from "@intx/types/runtime";
+} from "./attachments";
+import { base64Decode } from "./base64";
+import type { MessageAttachment } from "./runtime";
+import type { AttachmentError } from "./sessions";
 
 /**
- * A single attachment as it arrives on the request body: a MIME type, the
- * base64-encoded bytes, and an optional filename.
+ * A single attachment as it arrives at a boundary: a MIME type, the bytes,
```
alexanderguy: This docstring was correctly updated for the widening, but its sibling on `validateAttachments` further down wasn't — that one still opens "Validate and decode **request-body** attachments against a policy", and the line about names "defaulted by request-body position" has the same problem. Worth fixing in the same pass.

alexanderguy (changes requested):
**This review was produced by an agent running the `pull-request-review` skill.** A second agent independently re-verified the findings and reproduced the two significant ones at runtime against the real decode path; where it contradicted the first pass, the corrected reading is what appears here. Build and test status is taken from CI rather than re-run locally.

Commit messages are clean — four verb-first subjects, no prefixes, `Closes INTR-559` correctly confined to the PR body, and the lockfile travels with its `package.json`. The authorisation story holds up too: the capability handle is resolved once from the agent's own grants and no mail-tool argument carries an address to swap, so there's no cross-agent read path.

Requesting changes on two decode-path defects, both inline. Each hands the agent wrong content with no error, which is the failure mode worth spending effort on.

**Reachability, stated up front so these aren't over-weighted.** Both need foreign-shaped mail, and `resolveInboundMailPolicy` defaults every author-controllable outcome to reject — only a clean signature with matching `From` is admitted. Reaching them takes an authored `inboundMailPolicy: { missing: "admit" }` opt-in, or a custom host wiring `HubTransport.deliver` directly. Not live in the default composition. Still worth fixing: both are silent-corruption classes, `InMemoryTransport.deliver` does accept raw external bytes with only presence checks on Message-ID/From/Date, and `fetchFull` already carries a branch anticipating mail from a plain mail client. Neither new test exercises a non-writer shape, so CI wouldn't catch either.

**One thing outside this diff, flagged because the branch makes it routine.** `mail-part-store.ts` commits every decoded part — attachments included — as blobs under `runs/<runId>/parts/` in the run's git substrate, and exports no delete or prune. The substrate is git, so history retains them regardless; expunging the mail doesn't remove the bytes. Pre-existing and untouched here, but agent-authored attachments make it a routine producer, so the retention question is worth someone's attention separately. (Path construction is safe — `sanitizePartName` plus the mandatory index prefix rules out traversal.)

Also worth knowing: `mail_read` on a part path has no size cap, and fetching part `"1"` returns the whole multipart wrapper — every attachment at once — base64-encoded into model context. The `size` field lets a model decide, but nothing enforces it.

One small thing with no line to attach to, since it sits outside the diff hunks: the attachment-name guard in `validateAttachments` rejects quotes and line breaks but accepts an empty name, testing only `!== undefined`. An empty name emits `filename=""`, which the reader's first regex can't match and the fallback captures as the literal two-character string `""` — which that same guard would then reject if an agent tried to forward it. Two-line fix. (The guard is otherwise sufficient against header injection; a backslash-terminated name is a conformance wrinkle our own reader recovers from, not a vulnerability.)

